PR #7, opened 2020-12-02
[TH2-846] gRPC retries.

Ilya-Loginov-Exactpro:


src/main/java/com/exactpro/th2/common/schema/grpc/configuration/GrpcRetryConfiguration.java
```diff
@@ -0,0 +1,40 @@
...
+
+    @Override
+    public long getDelay(int currentAttempt) {
+        return minMethodRetriesTimeout + (maxRetriesAttempts > 1 ? (maxMethodRetriesTimeout - minMethodRetriesTimeout) / (maxRetriesAttempts - 1) * currentAttempt : 0);
+    }
+
+    @Override
+    public int getMaxAttempts() {
+        return maxRetriesAttempts;
+    }
```
OptimumOpium, line 39: Maybe we don't need to use a Lombok and generate getters and setters by ourselves? We will be able to have a field called `maxAttempts` and getters and setters for it. The getter will implement the `RetryPolicy` method.

src/main/java/com/exactpro/th2/common/schema/grpc/router/impl/DefaultGrpcRouter.java
```diff
@@ -61,7 +65,21 @@
...
+        TH2Impl implementationAnnotation = cls.getAnnotation(TH2Impl.class);
+
+        if (implementationAnnotation == null) {
+            return getProxyService(cls);
+        }
+
+        Class<?> th2ImplClass = implementationAnnotation.value();
+
+        try {
+            return (T) th2ImplClass.getConstructor(RetryPolicy.class, StubStorage.class).newInstance(configuration.getRetryConfiguration(), stubsStorages.computeIfAbsent(cls, key -> new DefaultStubStorage<>(getServiceConfig(key))));
```
OptimumOpium, line 79: I think the line is too long. We should split it into several separate lines (extract variables or call methods in chain each one on the new line)

src/main/java/com/exactpro/th2/common/schema/grpc/router/impl/DefaultGrpcRouter.java
```diff
@@ -61,7 +65,21 @@
...
+        } catch (InstantiationException | IllegalAccessException | InvocationTargetException | NoSuchMethodException e) {
+            throw new IllegalStateException("Can not create new instance of service", e);
```
OptimumOpium, line 81: Please, add information about the class we were trying to create an instance

src/main/java/com/exactpro/th2/common/schema/grpc/router/impl/DefaultStubStorage.java
```diff
@@ -0,0 +1,60 @@
...
+    @NotNull
+    @Override
+    public T getStub(@NotNull Message message, @NotNull AbstractStub.StubFactory<T> stubFactory) {
+        String endpointLabel = serviceConfiguration.getStrategy().getEndpoint(message);
+        return stubs.computeIfAbsent(endpointLabel, key -> {
+            GrpcEndpointConfiguration endpoint = serviceConfiguration.getEndpoints().get(key);
+
+            if (Objects.isNull(endpoint)) {
+                throw new IllegalStateException("No endpoint in configuration " +
+                        "that matching the provided alias: " + key);
```
OptimumOpium, line 54:
```suggestion
                throw new IllegalStateException("No endpoint in the configuration " +
                        "that matches the provided alias: " + key);
```

src/main/java/com/exactpro/th2/common/schema/grpc/router/impl/DefaultGrpcRouter.java
```diff
@@ -51,6 +54,7 @@
 
     private Map<Class<?>, Map<String, AbstractStub<?>>> stubs = new ConcurrentHashMap<>();
     private Map<String, Channel> channels = new ConcurrentHashMap<>();
+    private Map<Class<?>, StubStorage<?>> stubsStorages = new ConcurrentHashMap<>();
```
OptimumOpium, line 57: Why these fields are not final? I think we should add the final modifier

src/main/java/com/exactpro/th2/common/schema/grpc/router/impl/DefaultGrpcRouter.java
```diff
@@ -61,7 +65,21 @@
...
-        return getProxyService(cls);
+        Objects.requireNonNull(cls, "Services class can not be null");
+
+        TH2Impl implementationAnnotation = cls.getAnnotation(TH2Impl.class);
+
+        if (implementationAnnotation == null) {
+            return getProxyService(cls);
+        }
+
+        Class<?> th2ImplClass = implementationAnnotation.value();
```
OptimumOpium, line 76: I think we need to check if it is `null`

src/main/java/com/exactpro/th2/common/schema/grpc/router/impl/DefaultGrpcRouter.java
```diff
@@ -110,14 +128,17 @@ protected void validateArgs(Object[] args) {
     }
 
     protected AbstractStub<?> getGrpcStubToSend(Class<?> proxyService, Message message) throws ClassNotFoundException {
-        var grpcStubAnn = proxyService.getAnnotation(GrpcStub.class);
+        var grpcStubAnnOld = proxyService.getAnnotation(GrpcStub.class);
+        var grpcStubAnnNew = proxyService.getAnnotation(com.exactpro.th2.service.generator.service.annotation.GrpcStub.class);
```
OptimumOpium, line 132: If we are migrating to the new annotation we probably should import it instead of the old one.

OptimumOpium (changes requested):

src/main/java/com/exactpro/th2/common/schema/grpc/configuration/GrpcRetryConfiguration.java
```diff
@@ -0,0 +1,37 @@
...
+
+import com.exactpro.th2.service.RetryPolicy;
+
+public class GrpcRetryConfiguration implements RetryPolicy {
+
+    private int maxRetriesAttempts = 5;
+
+    private int minMethodRetriesTimeout = 100;
+
+    private int maxMethodRetriesTimeout = 2_000;
```
OptimumOpium, line 26: Why there are no setters for those fields? Reflection is good (sometimes...) but what about testing? How are you going to test this configuration?

src/main/java/com/exactpro/th2/common/schema/grpc/router/impl/DefaultGrpcRouter.java
OptimumOpium: Up. The class can be `null` (nobody will keep the users from setting it to null - we are not in Kotlin).

src/main/java/com/exactpro/th2/common/schema/grpc/router/impl/DefaultGrpcRouter.java
```diff
@@ -61,7 +65,26 @@
...
+
+        Class<?> th2ImplClass = implementationAnnotation.value();
+
+
+        try {
+            return (T) th2ImplClass.getConstructor(RetryPolicy.class, StubStorage.class)
+                    .newInstance(
+                            configuration.getRetryConfiguration(),
+                            stubsStorages.computeIfAbsent(cls, key ->
+                                    new DefaultStubStorage<>(getServiceConfig(key))));
```
OptimumOpium, line 84: I think we need to move closing brackets to the same level as the opening ones

src/main/java/com/exactpro/th2/common/schema/grpc/router/impl/DefaultStubStorage.java
```diff
@@ -0,0 +1,60 @@
...
+import io.grpc.stub.AbstractStub;
+
+@ThreadSafe
+public class DefaultStubStorage<T extends AbstractStub<T>> implements StubStorage<T> {
+
+    private final GrpcServiceConfiguration serviceConfiguration;
+    private final Map<String, T> stubs = new ConcurrentHashMap<>();
+
+    public DefaultStubStorage(GrpcServiceConfiguration serviceConfiguration) {
+        this.serviceConfiguration = serviceConfiguration;
```
OptimumOpium, line 42: We need to check if the passed configuration is not `null`

OptimumOpium (changes requested):

build.gradle
```diff
@@ -130,6 +130,9 @@ dependencies {
     api platform('com.exactpro.th2:bom:2.3.+')
     api 'com.exactpro.th2:cradle-core'
     api 'com.exactpro.th2:grpc-common'
+    
+    //FIXME: Move to grpc-common
+    implementation "com.exactpro.th2:service-generator-protoc:3.0.0"
```
OptimumOpium, line 135: Correct the artifact name

build.gradle
```diff
@@ -130,6 +130,9 @@ dependencies {
...
 
     implementation 'com.google.protobuf:protobuf-java-util'
     implementation 'com.exactpro.th2:grpc-service-generator'
```
OptimumOpium, line 138: It is the same dependency. We should use it instead of the dependency in line 135 or remove it

src/main/java/com/exactpro/th2/common/schema/grpc/router/impl/DefaultGrpcRouter.java
```diff
@@ -61,7 +65,28 @@
      * @throws ClassNotFoundException if matching the service class to protobuf stub has failed
      */
     public <T> T getService(@NotNull Class<T> cls) throws ClassNotFoundException {
-        return getProxyService(cls);
+        Objects.requireNonNull(cls, "Services class can not be null");
+
+        TH2Impl implementationAnnotation = cls.getAnnotation(TH2Impl.class);
+
+        if (implementationAnnotation == null || implementationAnnotation.value() == null) {
```
OptimumOpium, line 72: I don't think that the TH2Impl annotation with the `null` value is a normal case. Probably, we should throw an exception in this case.
If the annotation is not present at all it is ok. But if it presents it must have a non-null value

OptimumOpium (changes requested):

src/main/java/com/exactpro/th2/common/schema/grpc/router/impl/DefaultGrpcRouter.java
```diff
@@ -61,7 +66,32 @@
      * @throws ClassNotFoundException if matching the service class to protobuf stub has failed
      */
     public <T> T getService(@NotNull Class<T> cls) throws ClassNotFoundException {
-        return getProxyService(cls);
+        var implementations = ServiceLoader.load(Objects.requireNonNull(cls, "Services class can not be null"))
```
OptimumOpium, line 69: Please, specify the result type explicitly. It is not obvious that the result collection has the following type
```java
List<ServiceLoader.Provider<T>>
```

src/main/java/com/exactpro/th2/common/schema/grpc/router/impl/DefaultGrpcRouter.java
```diff
@@ -61,7 +66,32 @@
...
+        if (implementations.size() > 1) {
+            throw new IllegalStateException("Can not choose implementation. Fount " + implementations.size() + " implementations");
+        }
+
+        if (implementations.isEmpty()) {
+            //FIXME: Remove in future releases
+            return getProxyService(cls);
+        }
+
+        Class<?> th2ImplClass = implementations.get(0).type();
```
OptimumOpium, line 81: If you change the generic type for Class we won't get any warnings when we create an instance. Currectly we have 'uncheckedCast' waring on line 85
```suggestion
        Class<? extends T> th2ImplClass = implementations.get(0).type();
```

OptimumOpium (changes requested):

OptimumOpium (approved):

PR #9, opened 2020-12-04
[TH2-1087] Add MessageRouterUtils

ivandruzhinin:


gradle.properties
```diff
@@ -1,4 +1,4 @@
-release_version = 2.5.1
+release_version=2.5.2
```
Nikita-Smirnov-Exactpro, line 1: Please update the minor version because your change extends public API.

PR #10, opened 2020-12-08
[TH2-1128] Migrate to th2 storage 2.4

Nikita-Smirnov-Exactpro:
* Defined version for bom and cradle explicitly

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -117,7 +117,7 @@ ext {
     protobufVersion     = '3.11.0'
 
     sailfishVersion     = '3.2-SNAPSHOT'
-    cradleVersion       = '1.2-SNAPSHOT'
+    cradleVersion       = '2.4.0'
 
     jacksonVersion      = "2.11.2"
 }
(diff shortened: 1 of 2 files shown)


build.gradle
```diff
@@ -107,7 +107,7 @@ allprojects {
                 desc = project.version
                 released  = new Date()
             }
-        }
+            }
```
OptimumOpium, line 110: How has it happened?)

build.gradle
```diff
@@ -127,13 +127,13 @@ test {
 }
 
 dependencies {
-    api platform('com.exactpro.th2:bom:2.3.+')
-    api 'com.exactpro.th2:cradle-core'
+    api platform('com.exactpro.th2:bom:2.9.1')
+    api 'com.exactpro.th2:cradle-core:2.4.0'
```
OptimumOpium, line 131: Maybe we should extract the cradle version to a variable?

OptimumOpium (approved):

build.gradle
Nikita-Smirnov-Exactpro: I think no because this project has only one artifact related to cradle

OptimumOpium: And what about `implementation 'com.exactpro.th2:cradle-cassandra:2.4.0'`?

Nikita-Smirnov-Exactpro: O, skipped it, thanks

PR #22, opened 2021-02-12
switching from bintray to github packages

57R1D3R:


Nikita-Smirnov-Exactpro:
We have been migrated to the sonatype repository.

PR #33, opened 2021-03-03
[TH2-1537][CradleViewer] There is no way to change pageSize and timeout parameters in schema version of the cradle-viewer

MrJonniK:


MrJonniK:
abandoned

PR #41, opened 2021-03-25
[TS-461] Add ability to edit messages and values.

Ilya-Loginov-Exactpro:


src/main/kotlin/com/exactpro/th2/common/value/ValueUtils.kt
```diff
@@ -17,33 +17,70 @@
...
 fun Value.getLong(): Long? = this.getString()?.toLongOrNull()
 fun Value.getDouble(): Double? = this.getString()?.toDoubleOrNull()
 fun Value.getBigInteger(): BigInteger? = this.getString()?.toBigIntegerOrNull()
 fun Value.getBigDecimal(): BigDecimal? = this.getString()?.toBigDecimalOrNull()
-fun Value.getMessage(): Message? = if (this.kindCase == MESSAGE_VALUE) this.messageValue else null
+fun Value.getMessage(): Message? = if (hasMessageValue()) this.messageValue else null
+fun Value.getList() : List<Value>? = if (hasListValue()) this.listValue.valuesList else null
 
-fun String.toValue(): Value = Value.newBuilder().setSimpleValue(this).build()
+fun Value.Builder.updateList(updateFunc: ListValue.Builder.() -> ListValueOrBuilder?) : Value.Builder = apply { if (hasListValue()) updateOrAddList(updateFunc) else throw NullPointerException("Can not find list value") }
```
ivandruzhinin, line 42: Maybe change it like this:
```suggestion
fun Value.Builder.updateList(updateFunc: ListValue.Builder.() -> ListValueOrBuilder?) : Value.Builder = apply { check(hasListValue()) { "Can not find list value" }; updateOrAddList(updateFunc) }
```

src/main/kotlin/com/exactpro/th2/common/value/ValueUtils.kt
```diff
@@ -86,3 +125,17 @@ fun IntArray.toValue(): Value = toTypedArray().toValue()
 fun LongArray.toValue(): Value = toTypedArray().toValue()
 fun FloatArray.toValue(): Value = toTypedArray().toValue()
 fun DoubleArray.toValue(): Value = toTypedArray().toValue()
+
+
+fun Iterator<*>.toListValue() : ListValue = listValue().also { builder -> while(hasNext()) { next()?.also { builder.addValues(it.toValue()) } } }.build()
```
ivandruzhinin, line 130:
```suggestion
fun Iterator<*>.toListValue(): ListValue = listValue().also { list -> 
    forEach { 
        it?.toValue().run(list::addValues)
    }
}.build()
```
Also - why do we ignore null-values?

gradle.properties
```diff
@@ -13,7 +13,7 @@
 # limitations under the License.
 #
 
-release_version=3.5.0
+release_version=3.5.1
```
Nikita-Smirnov-Exactpro, line 16: Please up minor version because this change provided new public API

src/main/kotlin/com/exactpro/th2/common/schema/message/MessageRouterUtils.kt
```diff
@@ -57,7 +58,15 @@ fun MessageRouter<EventBatch>.storeEvent(
...
+fun MessageOrBuilder.toJson() : String = this.toJson(true)
+
+fun MessageOrBuilder.toJson(short: Boolean): String = JsonFormat.printer().includingDefaultValueFields().let {
+    (if (short) it.omittingInsignificantWhitespace() else it).print(this)
+}
+
+fun <T: Message.Builder> T.fromJson(json: String) : T {
+    JsonFormat.parser().ignoringUnknownFields().merge(json, this)
+    return this
+}
```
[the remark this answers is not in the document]
Nikita-Smirnov-Exactpro: I agree.

src/main/kotlin/com/exactpro/th2/common/value/ValueUtils.kt
```diff
@@ -17,33 +17,70 @@
...
+fun Value.Builder.updateList(updateFunc: ListValue.Builder.() -> ListValueOrBuilder?) : Value.Builder = apply { if (hasListValue()) updateOrAddList(updateFunc) else throw NullPointerException("Can not find list value") }
+fun Value.Builder.updateValue(updateFunc: String.() -> String?) : ValueOrBuilder = apply { simpleValue = updateFunc(simpleValue ?: throw NullPointerException("Can not find simple value")) }
+fun Value.Builder.updateMessage(updateFunc: Message.Builder.() -> MessageOrBuilder?) : Value.Builder = apply { if(hasMessageValue()) updateOrAddMessage(updateFunc) else throw NullPointerException("Can not find message value") }
 
-fun Message.toValue(): Value = Value.newBuilder().setMessageValue(this).build()
+fun Value.Builder.updateOrAddList(updateFunc: ListValue.Builder.() -> ListValueOrBuilder?) : Value.Builder = apply { updateFunc(listValueBuilder)?.also {
+    when (it) {
+        is ListValue -> listValue = it
+        is ListValue.Builder -> setListValue(it)
+        else -> throw IllegalStateException("Can not set list value")
```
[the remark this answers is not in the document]
Ilya-Loginov-Exactpro: Why we should print it?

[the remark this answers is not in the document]
ivandruzhinin: So it would be easier to understand what caused this `IllegalStateException`

src/main/kotlin/com/exactpro/th2/common/value/ValueUtils.kt
Ilya-Loginov-Exactpro: This behavior was in prev version. And null values do not have any useful information. If user want to add null value, he can use `add` method.

Nikita-Smirnov-Exactpro:
@Ilya-Loginov-Exactpro please add several tests for new functionality.

src/main/kotlin/com/exactpro/th2/common/schema/message/MessageRouterUtils.kt
```diff
@@ -56,17 +53,4 @@ fun MessageRouter<EventBatch>.storeEvent(
     }
 
     storeEvent(this, parentId)
-}
-
-// Not using default parameters for java compatibility
-@Deprecated(message = "Only for back compatibility", replaceWith = ReplaceWith("toJson(true)", imports = ["com.exactpro.th2.common.schema.message.toJson"]), level = DeprecationLevel.WARNING)
-fun MessageOrBuilder.toJson() : String = this.toJson(true)
```
ivandruzhinin: We should keep, but deprecate this function instead, to avoid breaking projects whom are already using it

src/main/kotlin/com/exactpro/th2/common/value/ValueUtils.kt
```diff
@@ -30,34 +30,34 @@ import java.math.BigInteger
...
+fun Value.Builder.updateValue(updateFunc: String.() -> String) : ValueOrBuilder = apply { simpleValue = updateFunc(simpleValue ?: throw NullPointerException("Can not find simple value")) }
+fun Value.Builder.updateMessage(updateFunc: Message.Builder.() -> MessageOrBuilder) : Value.Builder = apply { check(!hasMessageValue()) { "Can not find message value" }; updateOrAddMessage(updateFunc) }
 
-fun Value.Builder.updateOrAddList(updateFunc: ListValue.Builder.() -> ListValueOrBuilder?) : Value.Builder = apply { updateFunc(listValueBuilder)?.also {
+fun Value.Builder.updateOrAddList(updateFunc: ListValue.Builder.() -> ListValueOrBuilder) : Value.Builder = apply { updateFunc(listValueBuilder).also {
     when (it) {
         is ListValue -> listValue = it
         is ListValue.Builder -> setListValue(it)
-        else -> throw IllegalStateException("Can not set list value")
+        else -> throw IllegalStateException("Can not set list value. Wrong type = ${it.javaClass.typeName}")
```
ivandruzhinin, line 50:
```suggestion
        else -> error("Can not set list value. Wrong type = ${it.javaClass.typeName}")
```

src/main/kotlin/com/exactpro/th2/common/value/ValueUtils.kt
```diff
@@ -30,34 +30,34 @@ import java.math.BigInteger
...
     }
 } }
-fun Value.Builder.updateOrAddMessage(updateFunc: Message.Builder.() -> MessageOrBuilder?) : Value.Builder = apply { updateFunc(messageValueBuilder)?.also {
+fun Value.Builder.updateOrAddMessage(updateFunc: Message.Builder.() -> MessageOrBuilder) : Value.Builder = apply { updateFunc(messageValueBuilder).also {
     when(it) {
         is Message -> messageValue = it
         is Message.Builder -> setMessageValue(it)
-        else -> throw IllegalStateException("Can not set message value")
+        else -> throw IllegalStateException("Can not set message value. Wrong type = ${it.javaClass.typeName}")
```
ivandruzhinin, line 57:
```suggestion
        else -> error("Can not set message value. Wrong type = ${it.javaClass.typeName}")
```

src/main/kotlin/com/exactpro/th2/common/value/ValueUtils.kt
ivandruzhinin: Also why `typeName` instead of `canonicalName` / `qualifiedName`? And it's better to use `::class.java` instead of `.javaClass` because it could lead to an unexpected result inside of companion-object (which is not the case here, but still)

ivandruzhinin (dismissed):

ivandruzhinin (approved):

PR #56, opened 2021-04-16
[th2-1493] tries to load log4j.properties files from sources in order…

Nikita-Smirnov-Exactpro:
…: '/var/th2/config', '/home/etc', configured path via cmd, default configuration

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-# th2 common library (Java)
+# th2 common library (Java) (2.15.0)
 
 ## Usage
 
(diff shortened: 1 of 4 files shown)


Ilya-Loginov-Exactpro (approved):

jean-michel-perli-sonsin (changes requested):
Hi, please see if you agree with the changes below and correct them accordingly.

README.md - 3
Line 70:
password - the required setting defines the password that will be used **for connecting** to Cassandra.

line 71: cradleInstanceName - this option defines a special identifier that divides data within one keyspace with **infra set** as the default value.

src/main/java/com/exactpro/th2/common/schema/factory/AbstractCommonFactory.java - 42
Line 144/151: Create factory with **non-default** implementations schema classes

Line 305/312: **If** the router for that type is not registered yet **, it throws** {@link IllegalArgumentException}

Line 420/427: return Java bean with custom configuration, or <b>NULL</b> if configuration **does not exists** and **cannot** call default constructor from java bean class

The rest of the text inside README.md and the texts presented at the code itself are ok. 
Please let me know in case of any questions.

PR #67, opened 2021-04-27
[TH2-1733] Fix errors with loading dictionary.

Ilya-Loginov-Exactpro:


src/test/resources/test_load_dictionaries/prometheus.json
```diff
@@ -0,0 +1,3 @@
+{
+  "enabled": false
```
IvanChadin, line 2: Is it not true by default?

IvanChadin (changes requested):

IvanChadin (approved):

PR #68, opened 2021-04-29
[th2-1863] added toProtoEvent(EventID) and toProtoEvents(EventID) ove…

Nikita-Smirnov-Exactpro:
…rloads into the Event class

### DIFF
--- a/src/main/java/com/exactpro/th2/common/event/Event.java
+++ b/src/main/java/com/exactpro/th2/common/event/Event.java
@@ -269,6 +269,7 @@ public class Event {
     /**
      * @deprecated prefer to use full object instead of part of them, use the {@link #collectSubEvents(List, EventID)} method
      */
+    @Deprecated
     protected List<com.exactpro.th2.common.grpc.Event> collectSubEvents(List<com.exactpro.th2.common.grpc.Event> protoEvents, @Nullable String parentID) throws JsonProcessingException {
         return collectSubEvents(protoEvents, toEventID(parentID));
     }
--- a/src/main/java/com/exactpro/th2/common/event/Event.java
+++ b/src/main/java/com/exactpro/th2/common/event/Event.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2020-2020 Exactpro (Exactpro Systems Limited)
+ * Copyright 2020-2021 Exactpro (Exactpro Systems Limited)
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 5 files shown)


jean-michel-perli-sonsin (changes requested):
Hello, please see if you agree with the small changes suggested below:
README.md 
line 40 - connectionCloseTimeout - the timeout in milliseconds for completing all the close-related operations, use -1 for infinity, **the** default value is set to 10000.

line 42 - The `th2_readiness` probe is set to false and publishers are blocked after a lost connection to RabbitMQ. The `th2_readiness` probe is reverted to true if the connection will be recovered during specified attempts otherwise the `th2_liveness` probe will be set **to** false.

The rest of the text is correct.

Thanks,

src/main/java/com/exactpro/th2/common/event/Event.java
```diff
@@ -205,11 +207,27 @@ public Event messageID(MessageID attachedMessageID) {
...
 
+    /**
+     * @deprecated prefer to use full object instead of part of them, use the {@link #toProtoEvent(EventID)} method
+     */
+    @Deprecated
     public com.exactpro.th2.common.grpc.Event toProtoEvent(@Nullable String parentID) throws JsonProcessingException {
+        return toProtoEvent(toEventID(parentID));
+    }
+
+    public com.exactpro.th2.common.grpc.Event toProtoEvent(@Nullable EventID parentID) throws JsonProcessingException {
```
OptimumOpium, line 230: This new method will break the compatibility. If the original code calls `toProtoEvent(String)` with `null` parameter after adding that method the code won't compile. The compile can't choose between those two methods.

If it is okay and we are ready to change some places in code if we update the common-j version then ok, let's submit those changes

src/main/java/com/exactpro/th2/common/event/Event.java
```diff
@@ -205,11 +207,27 @@ public Event messageID(MessageID attachedMessageID) {
         return this;
     }
 
-    public List<com.exactpro.th2.common.grpc.Event> toProtoEvents(String parentID) throws JsonProcessingException {
+    /**
+     * @deprecated prefer to use full object instead of part of them, use the {@link #toListProto(EventID)} method
+     */
+    public List<com.exactpro.th2.common.grpc.Event> toProtoEvents(@Nullable String parentID) throws JsonProcessingException {
```
OptimumOpium, line 213: Missed `@Deprecated` annotation

src/main/java/com/exactpro/th2/common/event/Event.java
```diff
@@ -247,10 +265,17 @@ public Instant getEndTimestamp() {
         return endTimestamp;
     }
 
+    /**
+     * @deprecated prefer to use full object instead of part of them, use the {@link #collectSubEvents(List, EventID)} method
+     */
     protected List<com.exactpro.th2.common.grpc.Event> collectSubEvents(List<com.exactpro.th2.common.grpc.Event> protoEvents, @Nullable String parentID) throws JsonProcessingException {
```
OptimumOpium, line 271: Missed `@Deprecated` annotation

OptimumOpium (dismissed):

OptimumOpium (approved):

PR #73, opened 2021-05-12
[th2-1493] resets embedded log4j configuration before configuring fro…

Nikita-Smirnov-Exactpro:
…m a file


src/main/java/com/exactpro/th2/common/schema/factory/AbstractCommonFactory.java
```diff
@@ -767,6 +768,7 @@ protected static void configureLogger(String... paths) {
                 .filter(Files::exists)
                 .findFirst()
                 .ifPresentOrElse(path -> {
+                            LogManager.resetConfiguration();
                             PropertyConfigurator.configure(path.toString());
```
OptimumOpium, line 772: Maybe while we are here we can add the watch functionality? We only need to change the `configure` to `configureAndWatch`

OptimumOpium (approved):

PR #78, opened 2021-06-07
[TH2-1962] Add ability to read new filter format in common-j.

Ilya-Loginov-Exactpro:


src/main/kotlin/com/exactpro/th2/common/util/MultiMapUtils.kt
```diff
@@ -0,0 +1,69 @@
...
+    }
+}
+
+fun <K, V> emptyMultiMap() : MultiValuedMap<K, V> = MultiMapUtils.newListValuedHashMap()
+
+class DeserializerMultiMap<T>(private val valueClass: Class<T>, private val keySupplier: (T) -> String, private val fieldKeyName: String? = null) {
+    fun deserializerMultiMap(parser: JsonParser, node: JsonNode, name: String, location: JsonLocation): MultiValuedMap<String, T> {
+        val result: MultiValuedMap<String, T> = MultiMapUtils.newListValuedHashMap()
+        when {
+            node.isArray -> {
```
ivandruzhinin, line 45: Redundant brackets

src/main/kotlin/com/exactpro/th2/common/util/MultiMapUtils.kt
```diff
@@ -0,0 +1,69 @@
...
+                node.forEach { element ->
+                    parser.codec.treeToValue(element, valueClass)?.also { obj ->
```
ivandruzhinin, line 47: `parser.codec` can be extracted into a variable

src/main/kotlin/com/exactpro/th2/common/schema/grpc/configuration/GrpcRawFilterStrategy.kt
```diff
@@ -16,14 +16,53 @@
...
+
+    @JsonGetter("metadata")
+    fun getJsonMetadata(): Collection<FieldFilterConfiguration> = metadata.values()
+
+    @JsonGetter("message")
+    fun getJsonMessage(): Collection<FieldFilterConfiguration> = message.values()
+
+}
+
+class GrpcRouterFilterConfigurationDeserializer : JsonDeserializer<GrpcRouterFilterConfiguration>() {
```
OptimumOpium, line 50: Maybe we can create a single deserializer for MultiValuedMap<String, FieldFilterConfiguration> and apply it to the fields using `@JsonDeserializer` annotation?
We won't need to register any modules in ObjectMapper. Also, we won't need to check if the key is present (the deserializer won't invoked at all)

src/main/kotlin/com/exactpro/th2/common/util/MultiMapUtils.kt
```diff
@@ -0,0 +1,79 @@
...
+fun <K, V> multiMapOf(list: List<Pair<K, V>>): MultiValuedMap<K, V> =
+    MultiMapUtils.newListValuedHashMap<K, V>().also { multiMap ->
+        list.forEach { (key, value) ->
+            multiMap.put(key, value)
+        }
+    }
+
+fun <K, V> emptyMultiMap(): MultiValuedMap<K, V> = MultiMapUtils.newListValuedHashMap()
+
+class DeserializerMultiMap<T>(
```
Nikita-Smirnov-Exactpro, line 43:
```suggestion
//FIXME: Replace to simple bean when the major version will be updated
class DeserializerMultiMap<T>(
```

Nikita-Smirnov-Exactpro (changes requested):

README.md
```diff
@@ -188,6 +188,10 @@ NOTES:
 
 ## Release notes
 
+### 3.17
```
OptimumOpium, line 191: Please, write a full version

README.md
```diff
@@ -188,6 +188,10 @@ NOTES:
...
+
++ Add support for the new filter's format
```
OptimumOpium, line 193: Please, also add information about supporting multiple filters. Maybe we should add an example to the readme as well

src/main/kotlin/com/exactpro/th2/common/util/MultiMapUtils.kt
```diff
@@ -0,0 +1,71 @@
...
+                val fullNode = if (field.value is ObjectNode) (field.value as ObjectNode).set(
+                    "fieldName",
+                    TextNode(fieldName)
+                ) else node
+
+                codec.treeToValue(fullNode, FieldFilterConfiguration::class.java)?.also { filter ->
+                    result.put(filter.fieldName, filter)
+                }
+            }
+            else -> throw JsonParseException(parser, "Can not deserialize MultiValuedMap", location)
```
OptimumOpium, line 65: Please, use DeserializationContext.report* methods to report the bad JSON definition

src/main/kotlin/com/exactpro/th2/common/util/MultiMapUtils.kt
```diff
@@ -0,0 +1,71 @@
...
+                    result.put(filter.fieldName, filter)
+                }
+            }
+            node.isObject -> node.fields().forEach { field ->
+                val fieldName = field.key
+
+                val fullNode = if (field.value is ObjectNode) (field.value as ObjectNode).set(
+                    "fieldName",
+                    TextNode(fieldName)
+                ) else node
```
OptimumOpium, line 59: I think we can create a class that matches the old filter and use that class to deserialize the node value. After that, we can convert that class to the new filter class.
In my opinion, it is a better option because using the string with field name might cause some errors during refactoring (we might forget to change it when we change the field name)

src/test/kotlin/com/exactpro/th2/common/schema/TestJsonConfiguration.kt
```diff
@@ -49,113 +46,246 @@ class TestJsonConfiguration {
...
+                    )
+                )
+            )
+        )
     }
 
-    private fun testJson(configuration: Any) {
-        OBJECT_MAPPER.writeValueAsString(configuration).also {
-            assertEquals(configuration, OBJECT_MAPPER.readValue(it, configuration::class.java))
+    private fun testSerializeAndDeserialize(configuration: Any) {
```
OptimumOpium, line 282: I still think that this kind of test does not check anything. If our bean can be serialized and after that deserialized does not mean that the JSON that is produced by infra-operator will be correctly deserialized.
To make sure that they will need to use them as a source for deserialization. It is still only my opinion and if you disagree we can leave it as it is for now

OptimumOpium (changes requested):

src/main/java/com/exactpro/th2/common/schema/filter/strategy/impl/AbstractFilterStrategy.java
```diff
@@ -55,9 +55,10 @@ public boolean verify(T message, List<? extends RouterFilter> routerFilters) {
 
 
     private boolean checkValues(Map<String, String> messageFields, MultiValuedMap<String, FieldFilterConfiguration> fieldFilters) {
-        return fieldFilters.keys().stream().allMatch(fieldName -> {
+        return fieldFilters.size() == 0 || fieldFilters.keys().stream().anyMatch(fieldName -> {
```
ivandruzhinin, line 58:
```suggestion
        return fieldFilters.isEmpty() || fieldFilters.keys().stream().anyMatch(fieldName -> {
```

src/main/java/com/exactpro/th2/common/schema/filter/strategy/impl/AbstractFilterStrategy.java
```diff
@@ -55,9 +55,10 @@ public boolean verify(T message, List<? extends RouterFilter> routerFilters) {
...
+            String messageValue = messageFields.get(fieldName);
             Collection<FieldFilterConfiguration> filters = fieldFilters.get(fieldName);
-            return filters.stream().anyMatch(filter -> checkValue(messageFields.get(fieldName), filter));
+            return filters.size() != 0 && filters.stream().allMatch(filter -> checkValue(messageValue, filter));
```
ivandruzhinin, line 61:
```suggestion
            return !filters.isEmpty() && filters.stream().allMatch(filter -> checkValue(messageValue, filter));
```

src/main/kotlin/com/exactpro/th2/common/schema/strategy/route/json/RoutingStrategyModule.kt
```diff
@@ -0,0 +1,28 @@
+/*
+ * Copyright 2020-2021 Exactpro (Exactpro Systems Limited)
```
ivandruzhinin, line 2:
```suggestion
 * Copyright 2021-2021 Exactpro (Exactpro Systems Limited)
```

ivandruzhinin (changes requested):

ivandruzhinin (approved):

OptimumOpium (approved):

Nikita-Smirnov-Exactpro (approved):

PR #82, opened 2021-06-09
[TH2-1997] Change the way channels stored in the connection manager

OptimumOpium:


src/main/java/com/exactpro/th2/common/schema/message/impl/rabbitmq/connection/ConnectionManager.java
```diff
@@ -251,27 +262,31 @@ public void close() {
     }
 
     public void basicPublish(String exchange, String routingKey, BasicProperties props, byte[] body) throws IOException {
-        Channel channel = this.channel.get();
-        waitForConnectionRecovery(channel);
-        channel.basicPublish(exchange, routingKey, props, body);
+        ChannelHolder holder = getChannelFor(PinId.forRoutingKey(routingKey));
+        holder.withLock(channel -> {
+            waitForConnectionRecovery(channel);
```
Nikita-Smirnov-Exactpro, line 267: Maybe encapsulate into ChannelHolder and call it before operations

OptimumOpium: Yes, we can do that because we do checking for recovery before every action with a channel. However, I would also make the `basicCancel` method private because it might be invoked without locking. That is why we need to leave the waiting for connection recovery in that method if it is public method

OptimumOpium: As I can see nobody uses this method outside the `ConnectionManager`. But I cannot check the usage in other boxes

Nikita-Smirnov-Exactpro (approved):

PR #101, opened 2021-07-27
Migrated the Event.disperseToBatches extension method to Event

Nikita-Smirnov-Exactpro:


src/main/java/com/exactpro/th2/common/event/Event.java
```diff
@@ -302,16 +337,82 @@ protected String formatStateException(String fieldName, Object value) {
         return fieldName + " in event '" + id + "' already sed with value '" + value + '\'';
     }
 
-    @NotNull
+    /**
+     * @deprecated use {@link #getAggregatedStatus} instead
+     */
+    @Deprecated
```
OptimumOpium, line 343: We should add `forRemoval` flag in annotation

src/main/java/com/exactpro/th2/common/event/Event.java
```diff
@@ -302,16 +337,82 @@ protected String formatStateException(String fieldName, Object value) {
...
     }
 
+    private void batch(int maxEventBatchContentSize, List<EventBatch> result, Map<EventID, List<com.exactpro.th2.common.grpc.Event>> eventGroups, EventID eventID) throws JsonProcessingException {
+        EventBatch.Builder builder = EventBatch.newBuilder().setParentEventId(eventID);
+
+        List<com.exactpro.th2.common.grpc.Event> events = Objects.requireNonNull(eventGroups.get(eventID),
+                "Neither of events refers to " + shortDebugString(eventID));
+
+        for (com.exactpro.th2.common.grpc.Event event : events) {
+            com.exactpro.th2.common.grpc.Event checkedEvent = checkAndRebuild(maxEventBatchContentSize, event);
```
OptimumOpium, line 365: You can use `var` instead of type

src/main/java/com/exactpro/th2/common/event/Event.java
```diff
@@ -302,16 +337,82 @@ protected String formatStateException(String fieldName, Object value) {
...
+        int contentSize = getContentSize(builder);
+        if(contentSize > maxEventBatchContentSize) {
+            throw new IllegalStateException("The smallest batch size exceeds the max event batch content size, max " + maxEventBatchContentSize + ", actual " + contentSize);
+        }
+
+        return builder.build();
+    }
+
+    private com.exactpro.th2.common.grpc.Event checkAndRebuild(int maxEventBatchContentSize, com.exactpro.th2.common.grpc.Event event) throws JsonProcessingException {
+        if (getContentSize(event) > maxEventBatchContentSize) {
```
OptimumOpium, line 398: You can extract `getContentSize` into a variable and reuse in error event

src/main/java/com/exactpro/th2/common/event/Event.java
```diff
@@ -266,6 +275,32 @@ public Event messageID(MessageID attachedMessageID) {
...
+    public List<EventBatch> toListBatchProto(int maxEventBatchContentSize, @Nullable EventID parentID) throws JsonProcessingException {
+        if (maxEventBatchContentSize <= 0) {
+            throw new IllegalArgumentException("'maxEventBatchContentSize' should be greater than zero, actual: " + maxEventBatchContentSize);
+        }
+
+        List<com.exactpro.th2.common.grpc.Event> events = toListProto(parentID);
+
+        List<EventBatch> result = new ArrayList<>();
+        Map<EventID, List<com.exactpro.th2.common.grpc.Event>> eventGroups = events.stream().collect(Collectors.groupingBy(com.exactpro.th2.common.grpc.Event::getParentId));
+        batch(maxEventBatchContentSize, result, eventGroups, parentID);
```
OptimumOpium, line 300: What will happen if the `null` parentID was passed? As far as I remember the default value for Event::getParentId will be used. So when we will call `batch` with null `parentID` parameter the exception will be thrown because `eventGroups` does not have group for null parent ID

src/main/java/com/exactpro/th2/common/event/Event.java
```diff
@@ -302,16 +343,93 @@ protected String formatStateException(String fieldName, Object value) {
...
-            return subEvents.stream().anyMatch(subEvent -> subEvent.getAggrigatedStatus() == Status.FAILED)
+            return subEvents.stream().anyMatch(subEvent -> subEvent.getAggregatedStatus() == Status.FAILED)
                     ? Status.FAILED
                     : Status.PASSED;
         }
         return Status.FAILED;
     }
 
+    private void batch(int maxEventBatchContentSize, List<EventBatch> result, Map<EventID, List<com.exactpro.th2.common.grpc.Event>> eventGroups, @Nullable EventID eventID) throws JsonProcessingException {
+        eventID = ObjectUtils.defaultIfNull(eventID, DEFAULT_EVENT_ID);
```
OptimumOpium, line 365: Can be replaces with `Objects.requireNonNullElse`.
Let's start using Java standard library methods)

src/main/java/com/exactpro/th2/common/event/Event.java
```diff
@@ -302,16 +343,93 @@ protected String formatStateException(String fieldName, Object value) {
...
+        EventBatch.Builder builder = setParentId(EventBatch.newBuilder(), eventID);
+
+        List<com.exactpro.th2.common.grpc.Event> events = Objects.requireNonNull(eventGroups.get(eventID),
+                eventID != DEFAULT_EVENT_ID
```
OptimumOpium, line 369: I would prefer to invert the condition to avoid `not` in it

src/main/java/com/exactpro/th2/common/event/Event.java
```diff
@@ -302,16 +343,93 @@ protected String formatStateException(String fieldName, Object value) {
...
+                        ? "Neither of events refers to " + shortDebugString(eventID)
+                        : "Neither of events is root event");
+
+        for (var protoEvent : events) {
+            var checkedProtoEvent = checkAndRebuild(maxEventBatchContentSize, protoEvent);
+
+            LOGGER.trace("Process {} {}", checkedProtoEvent.getName(), checkedProtoEvent.getType());
+            if(eventGroups.containsKey(checkedProtoEvent.getId())) {
```
OptimumOpium, line 377: I wonder what will happen if the event won't have an ID set. I think we will get infinite recursion. But that should be a problem I hope

OptimumOpium (dismissed):

src/main/java/com/exactpro/th2/common/event/Event.java
```diff
@@ -266,6 +277,36 @@ public Event messageID(MessageID attachedMessageID) {
...
+            .addAllEvents(events);
+
+        if (parentID != null && events.size() != 1) {
+            builder.setParentEventId(parentID);
+        }
+
+        return builder.build();
+    }
+
+    public List<EventBatch> toListBatchProto(int maxEventBatchContentSize, @Nullable EventID parentID) throws JsonProcessingException {
```
OptimumOpium, line 293: Maybe rename it to `toBatchesProtoWithLimit`?

And also it would be great to have a JavaDoc

src/main/java/com/exactpro/th2/common/event/Event.java
Nikita-Smirnov-Exactpro: Added test

OptimumOpium (approved):

PR #106, opened 2021-08-06
[TH2-2189] Add ability to load secrets from a file

ivandruzhinin:


src/main/kotlin/com/exactpro/th2/common/schema/configuration/Secrets.kt
```diff
@@ -0,0 +1,13 @@
+package com.exactpro.th2.common.schema.configuration
```
OptimumOpium, line 1: Missing copyright

OptimumOpium (commented):
As far as I remember there should be an option to load the box configuration from the k8s. Do we need to add the secrets support for that mode as well?

ivandruzhinin:
> As far as I remember there should be an option to load the box configuration from the k8s. Do we need to add the secrets support for that mode as well?

I think this would handle that - https://github.com/th2-net/th2-common-j/commit/f762ee88b0593ec9864cc48d01c879dc058abc11#diff-6313585584b95b2851b150bc62e0bb207697681044acd9c28c1afa25246daf1dR480

OptimumOpium:
> > As far as I remember there should be an option to load the box configuration from the k8s. Do we need to add the secrets support for that mode as well?
> 
> I think this would handle that - [f762ee8#diff-6313585584b95b2851b150bc62e0bb207697681044acd9c28c1afa25246daf1dR480](https://github.com/th2-net/th2-common-j/commit/f762ee88b0593ec9864cc48d01c879dc058abc11#diff-6313585584b95b2851b150bc62e0bb207697681044acd9c28c1afa25246daf1dR480)

Yes, you are probably right. Missed that

PR #110, opened 2021-08-10
[TH2-1707] Change class of filter operation in router filters. Add operation 'WILDCARD'.

Ilya-Loginov-Exactpro:


src/main/java/com/exactpro/th2/common/schema/filter/strategy/impl/AbstractFilterStrategy.java
```diff
@@ -62,22 +63,24 @@ private boolean checkValues(Map<String, String> messageFields, MultiValuedMap<St
         });
     }
 
-    private boolean checkValue(String value1, FieldFilterConfiguration filterConfiguration) {
-        if (StringUtils.isEmpty(value1)) {
+    private boolean checkValue(String value, FieldFilterConfiguration filterConfiguration) {
+        if (StringUtils.isEmpty(value)) {
             return false;
```
OptimumOpium, line 68: Why are we returning `false` if the value is empty? What about the EMPTY filter?

src/main/java/com/exactpro/th2/common/schema/filter/strategy/impl/AbstractFilterStrategy.java
```diff
@@ -62,22 +63,24 @@ private boolean checkValues(Map<String, String> messageFields, MultiValuedMap<St
...
         }
 
-        var value2 = filterConfiguration.getExpectedValue();
+        var valueInConf = filterConfiguration.getExpectedValue();
 
         switch (filterConfiguration.getOperation()) {
```
OptimumOpium, line 73: Please, leave a comment about migrating to switch-expression when migrating to a newer java

src/main/java/com/exactpro/th2/common/schema/filter/strategy/impl/AbstractFilterStrategy.java
```diff
@@ -64,12 +64,9 @@ private boolean checkValues(Map<String, String> messageFields, MultiValuedMap<St
...
-        if (StringUtils.isEmpty(value)) {
-            return false;
-        }
-
         var valueInConf = filterConfiguration.getExpectedValue();
 
+        // FIXME: Change switch to switch-expression after upping java version
         switch (filterConfiguration.getOperation()) {
             case EQUAL:
                 return value.equals(valueInConf);
```
OptimumOpium, line 72: Shouldn't we check the value for `null`? Probably, we should use save equals (`Objects.equals(value, valueInConf)`)

OptimumOpium (commented):
Everything looks fine. But I don't see any block in the README file that describes how the user can use the filter for pins. Could you add such a block to the README?

OptimumOpium (approved):

PR #114, opened 2021-08-20
[TH2-2130] Added util to convert RootMessageFilter into readable collection of payload bodies

yarchiksmith:


src/test/kotlin/com/exactpro/th2/common/message/TestMessageFilterUtils.kt
```diff
@@ -84,6 +104,27 @@ class TestMessageFilterUtils {
...
+                putPropertyFilters("propA", simplePropertyFilter("valA", NOT_EQUAL, true))
+                putPropertyFilters("propB", simplePropertyFilter("valB"))
+            }
+            comparisonSettingsBuilder.apply {
+                addIgnoreFields("fieldA")
+                addIgnoreFields("fieldB")
+            }
+        }.build().toReadableBodyCollection(additionalMetadata)
+        Assertions.assertNotNull(toTreeTable)
+        Assertions.assertEquals(expected, objectMapper.writeValueAsString(toTreeTable))
```
OptimumOpium, line 125: As far as I remember there is no guarantee that the fields will be serialized in the same order on every machine. So, probably we need to serialize everything into **JsonNode**s and compare them

src/main/java/com/exactpro/th2/common/event/bean/TreeTable.java
```diff
@@ -16,13 +16,16 @@
...
+import com.fasterxml.jackson.annotation.JsonInclude;
 
 import java.util.Map;
 
+@JsonInclude(JsonInclude.Include.NON_NULL)
 public class TreeTable implements IBodyData {
     public static final String TYPE = "treeTable";
 
     private final String type = TYPE;
+    private String name;
```
OptimumOpium, line 28: I think we should make it final and set in the constructor. Also, don't forget to keep the current constructor for backward compatibility

OptimumOpium: Also, please check if the actual name is not blank (if it is `null` it is ok)

OptimumOpium (changes requested):

src/main/kotlin/com/exactpro/th2/common/message/MessageFilterUtils.kt
```diff
@@ -76,6 +82,28 @@ fun RootMessageFilter.toTreeTable(): TreeTable = TreeTableBuilder().apply {
...
+        }.build(),
+        MessageBuilder().apply {
+            text("Metadata")
+        }.build(),
+        TableBuilder<MetadataField>().apply {
+            row(MetadataField("message-type", messageType))
+            additionalMetadata?.forEach {
+                row(MetadataField(it.key, it.value))
+            }
+        }.build()
```
OptimumOpium, line 104: Can we use a tree table here as well? There won't be any difference in the result view except we will be able to set the header

OptimumOpium: Also, I think the information about the message type should go first (before filter and its settings)

src/test/kotlin/com/exactpro/th2/common/message/TestMessageFilterUtils.kt
```diff
@@ -84,6 +103,32 @@ class TestMessageFilterUtils {
...
+            comparisonSettingsBuilder.apply {
+                addIgnoreFields("fieldA")
+                addIgnoreFields("fieldB")
+            }
+        }.build().toReadableBodyCollection(additionalMetadata)
+
+        val expected = objectMapper.readTree(expectedJson).toList()
+        val actual = objectMapper.valueToTree<JsonNode>(toTreeTable).toList()
+
+        Assertions.assertNotNull(toTreeTable)
```
OptimumOpium, line 127: I think we should check it before calling `objectMapper.valueToTree` on line 125

src/test/kotlin/com/exactpro/th2/common/message/TestMessageFilterUtils.kt
```diff
@@ -84,6 +103,32 @@ class TestMessageFilterUtils {
...
+                putPropertyFilters("propB", simplePropertyFilter("valB"))
+            }
+            comparisonSettingsBuilder.apply {
+                addIgnoreFields("fieldA")
+                addIgnoreFields("fieldB")
+            }
+        }.build().toReadableBodyCollection(additionalMetadata)
+
+        val expected = objectMapper.readTree(expectedJson).toList()
+        val actual = objectMapper.valueToTree<JsonNode>(toTreeTable).toList()
```
OptimumOpium, line 125: Why can we simply compare two JsonNode? Why do we need to call `toList()`

src/main/java/com/exactpro/th2/common/event/bean/builder/TreeTableBuilder.java
```diff
@@ -26,14 +26,23 @@
     public static final String TABLE_TYPE = "treeTable";
```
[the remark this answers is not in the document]
yarchiksmith: I kept this field cause I was worried about backward compatibility

src/test/kotlin/com/exactpro/th2/common/message/TestMessageFilterUtils.kt
yarchiksmith: Our JsonNode is a list of serialized TreeTable. Two list elements are only compared as equal if they have the same values in the exact same order. In our case, the expected list has `Metadata` as the last element in the list and in the actual list as the first element. So, to compare two unordered lists, we have to convert the JsonNode to a list and check their size and elements

yarchiksmith: If we just compare two JsonNodes, we get failed check because the elements are unordered

src/main/java/com/exactpro/th2/common/event/bean/builder/TreeTableBuilder.java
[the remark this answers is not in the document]
OptimumOpium: You were right. All I meant is replace the `"treeTable"` with `TreeTable.TYPE`

```java
public static final String TABLE_TYPE = TreeTable.TYPE;
```

src/main/java/com/exactpro/th2/common/event/Event.java
```diff
@@ -207,6 +208,15 @@ public Event bodyData(IBodyData bodyData) {
         return this;
     }
 
+    /**
+     * Adds passed collection of body data
+     * @return current event
+     */
+    public Event bodyData(Collection<IBodyData> bodyDataCollection) {
```
OptimumOpium, line 215: Otherwise, we won't be able to add a collection with a certain type. E.g.
```java
class Body implements IBodyData {}

List<IBodyData> list = List.of(new Body());
new Event().bodyData(list); // works fine

List<Body> list = List.of(new Body());
new Event().bodyData(list); // compilation error
```
To fix the error we should change the method signature
```suggestion
    public Event bodyData(Collection<? extends IBodyData> bodyDataCollection) {
```

OptimumOpium (changes requested):

README.md
```diff
@@ -1,4 +1,4 @@
-# th2 common library (Java) (3.24.1)
+# th2 common library (Java) (3.25.9)
```
OptimumOpium, line 1:
```suggestion
# th2 common library (Java) (3.25.0)
```

OptimumOpium (changes requested):

OptimumOpium (approved):

PR #126, opened 2021-09-27
[TH2-2339] Add SimpleList to TreeTableEntry

Sergey-Sitnikov-Exactpro:


src/main/kotlin/com/exactpro/th2/common/message/MessageFilterUtils.kt
```diff
@@ -134,9 +134,12 @@ private fun ListValueFilter.toTreeTableEntry(): TreeTableEntry = CollectionBuild
...
 }.build()
 
-private fun SimpleFilter.toTreeTableEntry(): TreeTableEntry = RowBuilder()
-    .column(MessageFilterTableColumn(value, operation.toString(), key))
-    .build()
+private fun SimpleFilter.toTreeTableEntry(): TreeTableEntry = when {
+    hasSimpleList() -> simpleList.toTreeTableEntry(operation, key)
+    else -> RowBuilder()
+        .column(MessageFilterTableColumn(value, operation.toString(), key))
+        .build()
```
OptimumOpium, line 141: I would prefer to have `else` case throwing an exception about unsupported type with `kind` information. The same could be done for `ValueFilter.toTreeTableEntry()` as well
```suggestion
    hasValue() -> RowBuilder()
        .column(MessageFilterTableColumn(value, operation.toString(), key))
        .build()
    else -> error("Unsupported simple filter filter value: $filterValueCase")
```

OptimumOpium (changes requested):

OptimumOpium (approved):

PR #127, opened 2021-09-28
[TH2-2339] Fix Simple and Value filters treeTable convertation

Sergey-Sitnikov-Exactpro:


src/main/kotlin/com/exactpro/th2/common/message/MessageFilterUtils.kt
```diff
@@ -138,7 +140,7 @@ private fun ListValueFilter.toTreeTableEntry(): TreeTableEntry = CollectionBuild
 
 private fun SimpleFilter.toTreeTableEntry(): TreeTableEntry = when {
     hasSimpleList() -> simpleList.toTreeTableEntry(operation, key)
-    filterValueCase == VALUE -> RowBuilder()
+    filterValueCase == VALUE || filterValueCase == FILTERVALUE_NOT_SET -> RowBuilder()
```
OptimumOpium, line 143: I would say we should check the operation instead. If the operation does not expect to have a value (EMPTY, NOT_EMPTY) we should add a `MessageFilterTableColumn` with an empty value

OptimumOpium (approved):

PR #130, opened 2021-10-06
[th2-2469] builders for parsed/raw messages and events

andrew-drobynin:


src/main/java/com/exactpro/th2/common/event/EventBuilder.java
```diff
@@ -50,8 +50,8 @@
 import com.google.protobuf.ByteString;
 import com.google.protobuf.Timestamp;
 
-public class Event {
-    private static final Logger LOGGER = LoggerFactory.getLogger(Event.class);
+public class EventBuilder {
```
Nikita-Smirnov-Exactpro, line 53: Please revert the class name to save backward compatibility

src/main/java/com/exactpro/th2/common/schema/factory/CommonFactory.java
```diff
@@ -218,6 +222,18 @@ protected Path getOldPathToDictionariesDir() {
     protected ConfigurationManager getConfigurationManager() {
         return configurationManager;
     }
+    
+    public EventBuilder getEventBuilder() {
```
Nikita-Smirnov-Exactpro, line 226: Please add two new interfaces: MessageFactory and EventFactory. Instances of these factories can be taken from the AbstractCommonFactory
MessageFactory must have methods: createRawMessage, createParsedMessage
EventFactory must have the same set of methods as static methods in the Event class

src/main/java/com/exactpro/th2/common/event/EventBuilder.java
Nikita-Smirnov-Exactpro: Or we can inherit the Event class from EventBuilder

src/main/java/com/exactpro/th2/common/event/EventBuilder.java
```diff
@@ -70,33 +70,33 @@
     protected String description;
     protected Status status = Status.PASSED;
 
-    protected Event(Instant startTimestamp, @Nullable Instant endTimestamp) {
+    protected EventBuilder(Instant startTimestamp, @Nullable Instant endTimestamp) {
```
Nikita-Smirnov-Exactpro, line 73: constructors of this class should receive an instance of the EventFactory class to create sub events

src/main/kotlin/com/exactpro/th2/common/message/MessageBuilder.kt
```diff
@@ -0,0 +1,105 @@
...
+import com.exactpro.th2.common.grpc.ConnectionID
+import com.exactpro.th2.common.grpc.EventID
+import com.exactpro.th2.common.grpc.MessageID
+import com.google.protobuf.Timestamp
+import java.time.Instant
+
+abstract class MessageBuilder<T> {
+    private var sessionAlias: String? = null
+    private var direction: Int? = null
+    private var sequence: Long? = null
```
Nikita-Smirnov-Exactpro, line 27: Please switch these properties to the val type and fill them in a Constructor

src/main/kotlin/com/exactpro/th2/common/message/MessageBuilder.kt
```diff
@@ -0,0 +1,105 @@
...
+    private var subSequenceIndex: Int? = null
+    private var subSequenceValue: Int? = null
```
Nikita-Smirnov-Exactpro, line 29: please use list to store value of sub sequence

src/main/kotlin/com/exactpro/th2/common/message/MessageBuilder.kt
```diff
@@ -0,0 +1,105 @@
...
+    private var timestamp: Instant? = null
```
Nikita-Smirnov-Exactpro, line 30: This field should be read-only too, please use current time or time passed to a constructor

src/main/kotlin/com/exactpro/th2/common/message/MessageBuilder.kt
```diff
@@ -0,0 +1,105 @@
...
+    protected var properties: Map<String, String>? = null
+    protected var protocol: String? = null
```
Nikita-Smirnov-Exactpro, line 32: These can be nullable

src/main/kotlin/com/exactpro/th2/common/message/MessageBuilder.kt
```diff
@@ -0,0 +1,105 @@
...
+ */
+package com.exactpro.th2.common.message
+
+import com.exactpro.th2.common.grpc.ConnectionID
+import com.exactpro.th2.common.grpc.EventID
+import com.exactpro.th2.common.grpc.MessageID
+import com.google.protobuf.Timestamp
+import java.time.Instant
+
+abstract class MessageBuilder<T> {
```
Nikita-Smirnov-Exactpro, line 24: Please try to implement type-safe-builders
https://kotlinlang.org/docs/type-safe-builders.html#full-definition-of-the-com-example-html-package

src/main/kotlin/com/exactpro/th2/common/message/ParsedMessageBuilder.kt
```diff
@@ -0,0 +1,60 @@
...
+
+    fun addNullField(field: String): ParsedMessageBuilder {
+        this.fields[field] = Value.newBuilder().setNullValue(NullValue.NULL_VALUE).build()
+        return this
+    }
+
+    fun addSimpleField(field: String, value: String): ParsedMessageBuilder {
+        this.fields[field] = Value.newBuilder().setSimpleValue(value).build()
+        return this
+    }
```
Nikita-Smirnov-Exactpro, line 59: We should cover four types of values:
simple, message
list of simple, list of message

Nikita-Smirnov-Exactpro (changes requested):

src/main/java/com/exactpro/th2/common/event/EventBuilder.java
andreydrobynin: I renamed it back

src/main/kotlin/com/exactpro/th2/common/message/ParsedMessageBuilder.kt
andreydrobynin: By common.proto` in th2-grpc-common it can be:
```
message Value {
  oneof kind {
    NullValue null_value = 1;
    string simple_value = 2;
    Message message_value = 3;
    ListValue list_value = 4;
  }
}
```
I added methods for `Message` and `ListValue`.

src/main/kotlin/com/exactpro/th2/common/message/MessageBuilder.kt
andreydrobynin:
```
com.exactpro.th2.common.grpc.MessageMetadata.Builder#setProtocol
com.exactpro.th2.common.grpc.RawMessageMetadata.Builder#setProtocol
```
throw NullPointerException in case protocol is null

src/main/kotlin/com/exactpro/th2/common/message/MessageBuilder.kt
Nikita-Smirnov-Exactpro: sessionAlias, direction, sequence variables should be not nullable

src/main/kotlin/com/exactpro/th2/common/message/MessageBuilder.kt
Nikita-Smirnov-Exactpro: sorry, my mistake. 
These can not be nullable

Nikita-Smirnov-Exactpro: The default value of protocol is an empty string
the properties variable is a private modifiable map with a public method to put

src/main/kotlin/com/exactpro/th2/common/message/ParsedMessageBuilder.kt
ivandruzhinin: `list[...]` doesn't seem to be much different from built-in `listOf(...)`, I would've prefer to use built-in collection builders where possible

src/main/java/com/exactpro/th2/common/event/EventBuilder.java
andreydrobynin: done

src/main/kotlin/com/exactpro/th2/common/message/ParsedMessageBuilder.kt
Nikita-Smirnov-Exactpro: Do we combine both approaches in one builder?
For example 
java:
`var msg = factory.createMessage("session alias")
 msg.metadata() // getter 
      .sequence(3425) // setter
      .direction(FIRST) // setter
 msg.body() // getter
      .put("a", "value")
      .put("B", factory.createMessage()
          .body()
              .put("c", "value"))
      ...
`

src/main/java/com/exactpro/th2/common/message/MessageBuilder.java
```diff
@@ -0,0 +1,25 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.exactpro.th2.common.message;
+
+public interface MessageBuilder<Builder extends MessageBuilder<Builder, MB, R>, MB extends MetadataBuilder<MB>, R> {
+    Builder setParentEventId(String id, String bookName);
```
OptimumOpium, line 20: I am not sure but maybe we should use EventID instead of these two parameters

src/main/kotlin/com/exactpro/th2/common/message/BuilderExtension.kt
```diff
@@ -0,0 +1,164 @@
...
+    val event = Event.start(eventFactory)
+        .bookName(bookName)
+        .status(Event.Status.PASSED)
+        .name("name")
+        .type("type")
+        .bodyData(EventUtils.createMessageBean("bodyData"))
+        .messageID(MessageID.newBuilder().build())
+        .toProto(null)
+    println(event.toJson(false))
+}
```
OptimumOpium, line 164: We should remove test methods

src/main/java/com/exactpro/th2/common/event/Event.java
```diff
@@ -69,34 +69,58 @@
...
     }
 
     /**
      * Creates event with current time as start
+     * @param eventFactory
      * @return new event
      */
+    public static Event start(IEventFactory eventFactory) {
+        return eventFactory.start();
+    }
```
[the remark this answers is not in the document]
andreydrobynin: @OptimumOpium as I remember @Nikita-Smirnov-Exactpro asked to save back compatibility.

src/main/java/com/exactpro/th2/common/event/Event.java
```diff
@@ -69,34 +69,58 @@
...
     }
 
     /**
      * Creates event with passed time as start
+     * @param eventFactory
      * @return new event
      */
+    public static Event from(Instant startTimestamp, IEventFactory eventFactory) {
+        return eventFactory.from(startTimestamp);
+    }
```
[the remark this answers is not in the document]
andreydrobynin: Here also back compatibility

src/main/java/com/exactpro/th2/common/schema/message/RabbitMqTest.java
```diff
@@ -0,0 +1,298 @@
+package com.exactpro.th2.common.schema.message;
```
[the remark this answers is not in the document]
andreydrobynin: Of course. It's just an example and it will be deleted.

src/main/kotlin/com/exactpro/th2/common/event/EventFactory.kt
```diff
@@ -0,0 +1,30 @@
...
+import java.time.Instant
+
+class EventFactory : IEventFactory {
+    override fun start() = Event(this)
+
+    override fun from(startTimestamp: Instant) = Event(startTimestamp, this)
+
+    override fun from(startTimestamp: Instant, endTimestamp: Instant?) = from(startTimestamp, endTimestamp, this)
+
+    override fun from(startTimestamp: Instant, endTimestamp: Instant?, eventFactory: IEventFactory) =
```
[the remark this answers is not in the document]
andreydrobynin: @Nikita-Smirnov-Exactpro asked to add it. As I remember he wanted to have builder interface with old implementation (`com.exactpro.th2.common.event.Event`)

src/main/kotlin/com/exactpro/th2/common/schema/configuration/ConfigurationManager.kt
```diff
@@ -40,7 +40,7 @@ class ConfigurationManager(private val configurationPath: Map<Class<*>, Path>) {
             }
 
             val sourceContent = String(Files.readAllBytes(configPath))
-            LOGGER.info("Configuration path {} source content {}", configPath, sourceContent)
+            LOGGER.info("Configuration path $configPath source content $sourceContent")
```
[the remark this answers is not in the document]
andreydrobynin: I don't think second option is better but I used it.

src/main/java/com/exactpro/th2/common/event/Event.java
```diff
@@ -60,6 +60,7 @@
     protected static final ThreadLocal<ObjectMapper> OBJECT_MAPPER = ThreadLocal.withInitial(() -> new ObjectMapper().setSerializationInclusion(NON_NULL));
 
     protected final String id = generateUUID();
+    protected String bookName;
```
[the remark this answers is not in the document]
andreydrobynin: @Nikita-Smirnov-Exactpro asked to have builder methods (e.g. `bookName()`, `endTimestamp()`)

src/main/java/com/exactpro/th2/common/event/EventUtils.java
```diff
@@ -38,11 +38,18 @@ public static Message createMessageBean(String text) {
 
     @Contract("null -> null; !null -> !null")
     public static @Nullable EventID toEventID(@Nullable String id) {
```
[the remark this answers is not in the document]
andreydrobynin: done

src/main/java/com/exactpro/th2/common/message/Test.java
```diff
@@ -0,0 +1,115 @@
...
+import com.exactpro.th2.common.event.Event;
+import com.exactpro.th2.common.event.EventUtils;
+import com.exactpro.th2.common.event.IEventFactory;
+import com.exactpro.th2.common.grpc.MessageID;
+import com.exactpro.th2.common.message.impl.ParsedMessageBuilderImpl;
+import com.exactpro.th2.common.message.impl.RawMessageBuilderImpl;
+import com.exactpro.th2.common.schema.factory.CommonFactory;
+import com.fasterxml.jackson.core.JsonProcessingException;
+
+public class Test {
```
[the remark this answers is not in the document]
andreydrobynin: Of course. It's just an example and it will be deleted.

src/main/java/com/exactpro/th2/common/message/MessageBuilder.java
andreydrobynin: As I know we don't want to have dependencies to grpc EventID.

src/main/kotlin/com/exactpro/th2/common/message/BuilderExtension.kt
andreydrobynin: Of course. It's just an example and it will be deleted.

src/main/java/com/exactpro/th2/common/event/Event.java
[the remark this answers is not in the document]
OptimumOpium: It is a new method. How do we save backward compatibility here?

src/main/kotlin/com/exactpro/th2/common/event/EventFactory.kt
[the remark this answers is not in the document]
OptimumOpium: I still don't understand why do we need a method in the factory that accepts the factory as one of the parameters. How is it connected to an old implementation?

src/main/kotlin/com/exactpro/th2/common/schema/configuration/ConfigurationManager.kt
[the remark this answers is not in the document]
OptimumOpium: The difference between those options is that the first one will create a string no matter what but the second one will create a string only when the INFO level is enabled in the logger. Also, the last option does all concatenations only if the corresponding logger level is enabled

src/main/java/com/exactpro/th2/common/event/Event.java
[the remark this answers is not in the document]
OptimumOpium: I understand why we need the `endTimestamp` setter. But why do we need a setter for `bookName`? Why shouldn't we set it once in the constructor?

src/main/kotlin/com/exactpro/th2/common/schema/configuration/ConfigurationManager.kt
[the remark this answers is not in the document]
andreydrobynin: I see. Thanks!

src/main/java/com/exactpro/th2/common/event/Event.java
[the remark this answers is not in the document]
andreydrobynin: Sorry, I meant we want to have a new method with same name as an old. I think @Nikita-Smirnov-Exactpro can clarify it correctly than me.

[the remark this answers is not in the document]
OptimumOpium: From my point of view, it looks a bit strange. Why do I have to write `Event.start(eventFactory)` instead of `eventFactory.start()` in my code?

src/test/kotlin/com/exactpro/th2/common/event/TestEvent.kt
```diff
@@ -39,23 +40,24 @@ class TestEvent {
     private val data = EventUtils.createMessageBean("0123456789".repeat(20))
     private val dataSize = MAPPER.writeValueAsBytes(listOf(data)).size
     private val bigData = EventUtils.createMessageBean("0123456789".repeat(30))
+    private val commonFactory = CommonFactory.createFromArguments("-c", "src/test/resources/test_load_dictionaries")
```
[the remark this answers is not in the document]
andreydrobynin: done

PR #131, opened 2021-10-15
[TH2-114] Added supporting precision comparison for decimal values and timestamps

yarchiksmith:


README.md
```diff
@@ -251,6 +251,10 @@ NOTES:
 
 ## Release notes
 
+### 3.26.5
++ Migrated `grpc-common` version from `3.7.0` to `3.8.0`
+    + Added `time_precision` and `decimal_precision` parameters to `RootComparisonSettings`
```
OptimumOpium, line 256: Don't forget to add precision into `RootComparisonSettings.toTreeTableEntry()`

src/main/kotlin/com/exactpro/th2/common/message/MessageFilterUtils.kt
```diff
@@ -129,6 +134,17 @@ private fun RootComparisonSettings.toTreeTableEntry(): TreeTableEntry = Collecti
                 .build())
         }
     }.build())
+    if (hasTimePrecision()) {
+        val formattedDuration = DurationFormatUtils.formatDuration(Durations.toMillis(timePrecision), DEFAULT_TIME_PRECISION_FORMAT)
+        row("time-precision", RowBuilder()
+            .column(IgnoreFieldColumn(DEFAULT_TIME_PRECISION_REGEX.replace(formattedDuration, "").trim()))
```
OptimumOpium, line 140: If you are willing to use regexp maybe it is better to use something like that instead of DurationFormatUtils
https://stackoverflow.com/a/40487511

PR #132, opened 2021-10-18
[th2-2401] added dependency to cradleapi 3.0.0-TH2-2401-1347174066-SNAPSHOT

andreydrobynin:


Nikita-Smirnov-Exactpro:
This PR is redundant because the 3.0.0 cradle API is experimental

PR #149, opened 2021-11-24
Fix printing of empty MessageGroupBatch in debug logs of MessageGroup…

ivandruzhinin:
…Batch router

### DIFF
--- a/src/main/kotlin/com/exactpro/th2/common/message/MessageUtils.kt
+++ b/src/main/kotlin/com/exactpro/th2/common/message/MessageUtils.kt
@@ -25,6 +25,8 @@ import com.exactpro.th2.common.event.bean.builder.CollectionBuilder
 import com.exactpro.th2.common.event.bean.builder.RowBuilder
 import com.exactpro.th2.common.event.bean.builder.TreeTableBuilder
 import com.exactpro.th2.common.grpc.AnyMessage
+import com.exactpro.th2.common.grpc.AnyMessage.KindCase.MESSAGE
+import com.exactpro.th2.common.grpc.AnyMessage.KindCase.RAW_MESSAGE
 import com.exactpro.th2.common.grpc.ConnectionID
 import com.exactpro.th2.common.grpc.Direction
 import com.exactpro.th2.common.grpc.FailUnexpected
@@ -298,11 +300,15 @@ val Message.direction
 var Message.Builder.direction
     get(): Direction = metadata.id.direction
     set(value) {
-        setMetadata(MessageMetadata.newBuilder(metadata).apply {
-            setId(MessageID.newBuilder(id).apply {
-                direction = value
-            })
-        })
+        metadataBuilder.idBuilder.direction = value
+    }
+
+val RawMessage.direction
+    get(): Direction = metadata.id.direction
+var RawMessage.Builder.direction
+    get(): Direction = metadata.id.direction
+    set(value) {
+        metadataBuilder.idBuilder.direction = value
     }
 
 val Message.sessionAlias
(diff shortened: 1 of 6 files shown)


src/main/kotlin/com/exactpro/th2/common/schema/message/MessageRouterUtils.kt
```diff
@@ -71,4 +76,35 @@ fun appendAttributes(
     return mutableSetOf(*attributes).apply {
         addAll(requiredAttributes())
     }
+}
+
+fun MessageGroupBatch.toShortDebugString(): String {
+    val sessionAliases = mutableListOf<String>()
+    val directions = mutableListOf<Direction>()
```
OptimumOpium, line 83: Why lists? We will get a lot of repeating values in the result if the batch contains all messages from the same session and direction

ivandruzhinin: Yep, you're right, my bad

ivandruzhinin: Maybe it would be better to print message id's instead?

OptimumOpium: Do you mean all message IDS? If that so, I think it will be better but only in case we will form the message ID ourselves and won't use `toJson` method.
Something like: <session>:<direction>:<sequence>[.<subsequences>+]

src/main/kotlin/com/exactpro/th2/common/message/MessageUtils.kt
```diff
@@ -298,37 +300,82 @@ val Message.direction
...
+            clearSubsequence()
+            addAllSubsequence(value)
+        }
+    }
+
+val Message.logId: String
+    get() = "$sessionAlias:${direction.toString().toLowerCase()}:$sequence${subsequence.joinToString("") { ":$it" }}"
+
+val RawMessage.logId: String
+    get() = "$sessionAlias:${direction.toString().toLowerCase()}:$sequence${subsequence.joinToString("") { ":$it" }}"
```
OptimumOpium, line 372: Provider will use dots to separate subsequences. I think we should use the same format
```suggestion
    get() = "$sessionAlias:${direction.toString().toLowerCase()}:$sequence${subsequence.joinToString("") { ".$it" }}"

val RawMessage.logId: String
    get() = "$sessionAlias:${direction.toString().toLowerCase()}:$sequence${subsequence.joinToString("") { ".$it" }}"
```

src/main/kotlin/com/exactpro/th2/common/schema/message/MessageRouterUtils.kt
```diff
@@ -71,4 +74,16 @@ fun appendAttributes(
...
     }
+}
+
+fun MessageGroupBatch.toShortDebugString(): String = buildString {
+    append("MessageGroupBatch(ids = ")
+
+    groupsList.asSequence()
+        .flatMap { it.messagesList.asSequence() }
+        .map(AnyMessage::logId)
+        .toSortedSet()
```
OptimumOpium, line 85: Why sorted set? To sort it by session alias? Because we will sort the long ID and I don't think that sorting will perform the way you expect

ivandruzhinin: > To sort it by session alias?

Yeah.

> I don't think that sorting will perform the way you expect

Why not?

OptimumOpium: My bad, the colon will always be lower than any letter. Will work fine, I guess.

src/main/kotlin/com/exactpro/th2/common/message/MessageUtils.kt
```diff
@@ -298,37 +300,82 @@ val Message.direction
...
+    get(): List<Int> = metadata.id.subsequenceList
+    set(value) {
+        metadataBuilder.idBuilder.apply {
+            clearSubsequence()
+            addAllSubsequence(value)
+        }
+    }
+
+val Message.logId: String
+    get() = "$sessionAlias.${direction.toString().toLowerCase()}.$sequence${subsequence.joinToString("") { ".$it" }}"
```
OptimumOpium, line 369: I meant only subsequences...the session, direction and sequence should joined with colon

ivandruzhinin: Sorry, missed the part about subsequences

ivandruzhinin: Also why did we decide to do it this way? I've always though of message-id as some kind of path and it would make sense that all of its parts would be separated by the same separator

OptimumOpium (approved):

PR #150, opened 2021-11-24
[TH2-2727] Add method for getting stub for all registered endpoints

OptimumCode:


src/main/java/com/exactpro/th2/common/schema/grpc/router/impl/DefaultGrpcRouter.java
```diff
@@ -77,32 +78,70 @@
...
+    public <T> Set<T> getServices(@NotNull Class<T> serviceClass) {
+        List<Class<? extends T>> implementationClasses = loadImplementations(serviceClass);
+        if (implementationClasses.size() > 1) {
+            throw new IllegalStateException("Can not choose implementation for " + serviceClass +
+                    ". Fount " + implementationClasses.size() + " implementations");
+        }
+
+        if (implementationClasses.isEmpty()) {
+            throw new IllegalStateException("Cannot find any implementation for " + serviceClass);
+        }
```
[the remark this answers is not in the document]
OptimumOpium: We can do it in the future. But currently, in the case when we load only one service there is a fallback to the old implementation. Or you are talking about creating this method and reusing it later in the `getService` method when we get rid of an old implementation?

src/main/java/com/exactpro/th2/common/schema/grpc/router/GrpcRouter.java
```diff
@@ -39,7 +41,17 @@
...
-    <T> T getService(@NotNull Class<T> cls) throws ClassNotFoundException;
+    <T> T getService(@NotNull Class<T> cls);
+
+    /**
+     * Returns the gRPC service for each endpoint specified for passed class
+     * @param <T> the type of gRPC service to create
+     * @param serviceClass the class of corresponding service
+     * @return the service for each endpoint specified in the configuration
+     */
+    default <T> Set<T> getServices(@NotNull Class<T> serviceClass) {
```
Nikita-Smirnov-Exactpro, line 52: Why do we use Set instead of List?

OptimumOpium: Because there should not be any dependencies on the order of elements. Also, it shows that each service is unique.

Nikita-Smirnov-Exactpro: We should hint the developer about the requirement to implement equal / hashCode methods in class T to guarantee uniqueness.

OptimumOpium: Should we? We generate those implementations by proto files. Otherwise, they won't work like common expects.
But if we are afraid that the developer will write a new implementation with unpredictable equal / hashCode I will change it to a List

OptimumCode:
This functionality is no longer required due to migration to a new approach in which the processor is responsible for requesting data

PR #155, opened 2021-12-01
[th2-2762] added prepareStorage property to cradle.json

andreydrobynin:
Part of https://github.com/th2-net/th2-common-j/pull/156

### DIFF
--- a/src/main/kotlin/com/exactpro/th2/common/schema/cradle/CradleConfiguration.kt
+++ b/src/main/kotlin/com/exactpro/th2/common/schema/cradle/CradleConfiguration.kt
@@ -31,8 +31,9 @@ data class CradleConfiguration(
     var timeout: Long,
     var pageSize: Int,
     var cradleMaxEventBatchSize: Long,
-    var cradleMaxMessageBatchSize: Long
-)  : Configuration() {
+    var cradleMaxMessageBatchSize: Long,
+    var prepareStorage: Boolean
+) : Configuration() {
     constructor(
         cradleConfidentialConfiguration: CradleConfidentialConfiguration,
         cradleNonConfidentialConfiguration: CradleNonConfidentialConfiguration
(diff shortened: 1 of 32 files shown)


src/main/kotlin/com/exactpro/th2/common/schema/cradle/CradleConfiguration.kt
```diff
@@ -65,5 +67,6 @@ data class CradleNonConfidentialConfiguration(
     var timeout: Long = CassandraStorageSettings.DEFAULT_TIMEOUT,
     var pageSize: Int = 5000,
     var cradleMaxEventBatchSize: Long = CassandraStorageSettings.DEFAULT_MAX_EVENT_BATCH_SIZE,
-    var cradleMaxMessageBatchSize: Long = CassandraStorageSettings.DEFAULT_MAX_MESSAGE_BATCH_SIZE
+    var cradleMaxMessageBatchSize: Long = CassandraStorageSettings.DEFAULT_MAX_MESSAGE_BATCH_SIZE,
+    var prepareStorage: Boolean = false
```
Nikita-Smirnov-Exactpro, line 71: @MrJonniK @std000 do we want to use the false value as default?

std000: > do we want to use the false value as default?

I'd think that storage should be prepared by our "admin tool" by default.
So this option should be false here.

MrJonniK: I think we should use "false" as default, since in this case it is enough to send "true" only to one box to initialize storage.

Nikita-Smirnov-Exactpro: I think we should not change the default behaviour of estore/mstore/rpt-data-provider. We should change it via the new version of infra which set false to the cradle_manager.json config by default.

PR #179, opened 2022-01-20
[TH2-3043] Manual delivery acknowledgement

OptimumOpium:


src/main/kotlin/com/exactpro/th2/common/schema/message/ConfirmationMessageListener.kt
```diff
@@ -0,0 +1,63 @@
...
+
+    @Throws(Exception::class)
+    fun handle(consumerTag: String, message: T, confirmation: ManualAckDeliveryCallback.Confirmation)
+
+    fun onClose() {}
+
+    companion object {
+        @JvmStatic
+        @JvmOverloads
+        fun <T> wrap(listener: MessageListener<T>, confirm: Boolean = false): ConfirmationMessageListener<T> = DelegateListener(listener, confirm)
```
Nikita-Smirnov-Exactpro, line 36: Maybe the confirm option is always true because we move the autoconfirm logic for the ConnectionManager and the listener don't know about ManualAckDeliveryCallback.Confirmation

src/main/java/com/exactpro/th2/common/schema/message/impl/rabbitmq/AbstractRabbitSubscriber.java
```diff
@@ -156,7 +159,13 @@ public void start() throws Exception {
     }
 
     @Override
-    public void addListener(MessageListener<T> messageListener) {
+    public void addListener(ConfirmationMessageListener<T> messageListener) {
+        if (messageListener.getManualConfirmation()) {
+            if (hasManualSubscriber.compareAndSet(false, true)) {
+                throw new IllegalStateException("cannot subscribe listener " + messageListener
+                        + " because only one listener with manual confirmation is allowed per queue");
```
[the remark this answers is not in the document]
Nikita-Smirnov-Exactpro: Also, users catch the exception about only one problem when manual subscription via the subscribeAllWithManualAck method overlaps several pins that already have auto subscribers.

Nikita-Smirnov-Exactpro (changes requested):

build.gradle
```diff
@@ -184,6 +184,7 @@ dependencies {
     implementation "com.fasterxml.jackson.core:jackson-annotations"
     implementation "com.fasterxml.jackson.datatype:jackson-datatype-jsr310"
     implementation "com.fasterxml.jackson.dataformat:jackson-dataformat-yaml"
+    // TODO: need to update jackson kotlin module to get rid of kotlin-reflect 1.3.72
```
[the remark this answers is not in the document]
OptimumOpium: Why? That is a reminder that we should do that. Using different versions of kotlin in one application might cause some strange behavior

src/main/java/com/exactpro/th2/common/schema/message/impl/rabbitmq/AbstractRabbitSubscriber.java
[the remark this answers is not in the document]
OptimumOpium: The information about pin is not available here. But it is handled in `AbstractRabbitRouter`.

Regarding the second comment: could you please be more specific? I really try to get what do you mean but can't

src/main/kotlin/com/exactpro/th2/common/schema/message/ConfirmationMessageListener.kt
OptimumOpium: In case there is more than one listener for one queue and one of them has manual acknowledgment the other ones should not confirm the delivery.  That is why the default value for `confirm` parameter is `false`

OptimumOpium: But maybe we should use `confirm` parameter as a value to return from `manualConfirmation` property

src/main/java/com/exactpro/th2/common/schema/message/impl/rabbitmq/AbstractRabbitSubscriber.java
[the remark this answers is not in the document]
Nikita-Smirnov-Exactpro: > Regarding the second comment: could you please be more specific? I really try to get what you mean but can't
we have the config
```
pins
  - name: in_client
    connection-type: mq
    attributes:
      - A
      - C
  - name: in_client
    connection-type: mq
    attributes:
      - A
      - D
```
router.subscribeWithManualAck(..., "C")
router.subscribeWithManualAck(..., "D")
router.subscribeAllWithManualAck(..., "A")

src/main/kotlin/com/exactpro/th2/common/schema/message/ConfirmationMessageListener.kt
Nikita-Smirnov-Exactpro: I didn't find a pace where the wrap method is called with the confirm argument

src/main/java/com/exactpro/th2/common/schema/message/impl/rabbitmq/connection/ConnectionManager.java
```diff
@@ -275,21 +275,17 @@ public void basicPublish(String exchange, String routingKey, BasicProperties pro
...
+        String tag = holder.mapWithLock(channel ->
+                channel.basicConsume(queue, false, subscriberName + "_" + nextSubscriberId.getAndIncrement(), (tagTmp, delivery) -> {
                     try {
-                        try {
-                            deliverCallback.handle(tagTmp, delivery);
-                        } finally {
-                            holder.withLock(ch -> basicAck(ch, delivery.getEnvelope().getDeliveryTag()));
-                        }
+                        deliverCallback.handle(tagTmp, delivery,
+                                () -> holder.withLock(ch -> basicAck(ch, delivery.getEnvelope().getDeliveryTag())));
```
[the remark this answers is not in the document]
Nikita-Smirnov-Exactpro: https://www.rabbitmq.com/confirms.html#consumer-acks-double-acking
```
Should a client acknowledge the same delivery tag more than once, RabbitMQ will result a channel error such as PRECONDITION_FAILED - unknown delivery tag 100. The same channel exception will be thrown if an unknown delivery tag is used.
```

src/main/java/com/exactpro/th2/common/schema/message/impl/rabbitmq/connection/ConnectionManager.java
```diff
@@ -265,23 +279,46 @@ public void close() {
...
+
+                        Confirmation confirmation = () -> holder.withLock(ch -> {
+                            try {
+                                basicAck(ch, deliveryTag);
+                            } finally {
+                                holder.release(() -> metrics.getReadinessMonitor().enable());
+                            }
+                        });
+
+                        holder.acquireAndSubmitCheck(() ->
```
Nikita-Smirnov-Exactpro, line 309: acquireAndSubmitCheck is called without lock, but it changes the holder's variable `check`

src/main/kotlin/com/exactpro/th2/common/schema/message/impl/rabbitmq/configuration/RabbitMQConfiguration.kt
```diff
@@ -36,5 +37,12 @@ data class ConnectionManagerConfiguration(
     var minConnectionRecoveryTimeout: Int = 10000,
     var maxConnectionRecoveryTimeout: Int = 60000,
     val prefetchCount: Int = 10,
-    val messageRecursionLimit: Int = 100
-) : Configuration()
\ No newline at end of file
+    val messageRecursionLimit: Int = 100,
+    val workingThreads: Int = 1,
+    val confirmationTimeout: Duration = Duration.ofMinutes(15)
```
Nikita-Smirnov-Exactpro, line 42:
```suggestion
    val confirmationTimeout: Duration = Duration.ofMinutes(5)
```

Nikita-Smirnov-Exactpro: Most of our boxes should handle batches as faster as possible because they are part of conveyor

Nikita-Smirnov-Exactpro (dismissed):

Nikita-Smirnov-Exactpro (dismissed):

Nikita-Smirnov-Exactpro (approved):

PR #181, opened 2022-01-30
[TH2-2269] Add setting to be able to change maxMessageSize

alexey-suknatov-exactpro:


isengrims:
Closed because duplicated in https://github.com/th2-net/th2-common-j/pull/230

PR #189, opened 2022-02-16
Th2 3202

Topru333:


src/main/java/com/exactpro/th2/common/schema/factory/AbstractCommonFactory.java
```diff
@@ -529,12 +529,21 @@ public CradleManager getCradleManager() {
      */
     public abstract InputStream loadSingleDictionary();
 
+    /**
+     * @param alias name of dictionary
+     * @return Dictionary as {@link File}
+     * @throws IllegalStateException if can not read dictionary
+     */
+    public abstract File loadDictionaryFile(String alias);
```
OptimumOpium, line 537: I think the name should be `getDicrionaryFile` or `getPathToDictionary`. And it is better to use `Path` as a result I think

src/main/java/com/exactpro/th2/common/schema/factory/CommonFactory.java
```diff
@@ -545,8 +545,22 @@ public InputStream loadSingleDictionary() {
         }
     }
 
+    @Override
+    public File loadDictionaryFile(String alias) {
```
OptimumOpium, line 549: The same here

OptimumOpium (commented):
I am not sure that there are any benefits from having a method that returns a path to the archived file. At least we should provide the interface for unzipping it

OptimumOpium:
> I am not sure that there are any benefits from having a method that returns a path to the archived file. At least we should provide the interface for unzipping it

I should highlight it once again because I honestly don't understand the reason why we can simply return the path to a file that is a ZIPed file. We either should perform unzipping by the common lib or somehow provide user with a simple way to unzip this file somewhere

@Nikita-Smirnov-Exactpro @ivandruzhinin @Topru333 what do you think about that?

Nikita-Smirnov-Exactpro:
>> I am not sure that there are any benefits from having a method that returns a path to the archived file. At least we should provide the interface for unzipping it

> I should highlight it once again because I honestly don't understand the reason why we can simply return the path to a file that is a ZIPed file. We either should perform unzipping by the common lib or somehow provide user with a simple way to unzip this file somewhere

@Nikita-Smirnov-Exactpro @ivandruzhinin @Topru333 what do you think about that?

@OptimumOpium I agree with your commen. Maybe unzip all dictionaries and save them in special directory at the first call related to dictionary access

Nikita-Smirnov-Exactpro:
> > > > I am not sure that there are any benefits from having a method that returns a path to the archived file. At least we should provide the interface for unzipping it
> > 
> > 
> > > I should highlight it once again because I honestly don't understand the reason why we can simply return the path to a file that is a ZIPed file. We either should perform unzipping by the common lib or somehow provide user with a simple way to unzip this file somewhere
> > 
> > 
> > @Nikita-Smirnov-Exactpro @ivandruzhinin @Topru333 what do you think about that?
> > @OptimumOpium I agree with your commen. Maybe unzip all dictionaries and save them in special directory at the first call related to dictionary access
> 
> Slightly unrelated question - why do we even serve zipped files to common? I know that we zip files due k8s limitations, but I don't see any reason to not unzip them when a box is deployed (but this probably would be breaking change though)

Kubernetes provide ability mount resources from configmap, image responsible for additional actions as I understand.

ivandruzhinin:
> > > > > I am not sure that there are any benefits from having a method that returns a path to the archived file. At least we should provide the interface for unzipping it
> > > 
> > > 
> > > > I should highlight it once again because I honestly don't understand the reason why we can simply return the path to a file that is a ZIPed file. We either should perform unzipping by the common lib or somehow provide user with a simple way to unzip this file somewhere
> > > 
> > > 
> > > @Nikita-Smirnov-Exactpro @ivandruzhinin @Topru333 what do you think about that?
> > > @OptimumOpium I agree with your commen. Maybe unzip all dictionaries and save them in special directory at the first call related to dictionary access
> > 
> > 
> > Slightly unrelated question - why do we even serve zipped files to common? I know that we zip files due k8s limitations, but I don't see any reason to not unzip them when a box is deployed (but this probably would be breaking change though)
> 
> Kubernetes provide ability mount resources from configmap, image responsible for additional actions as I understand.

Then we should probably unzip dictionaries on demand and serve their path or content when required

PR #205, opened 2022-05-27
Enabled grpc server reflection

Nikita-Smirnov-Exactpro:


OptimumOpium (changes requested):
I think we should also add a small notice to the readme.
And update the version

README.md
```diff
@@ -239,6 +239,17 @@ The default attributes are:
 
 This library allows you to:
 
+## Routers
+
+### gRPC router
+
+This kind of router provides ability to interact boxes to interact between each other via gRPC interface.
```
jean-michel-perli-sonsin, line 246:
```suggestion
This kind of router provides the ability for boxes to interact between each other via gRPC interface.
```

README.md
```diff
@@ -239,6 +239,17 @@ The default attributes are:
...
+
+#### Server
+
+gRPC router rises a gRPC server with enabled [grpc-service-reflection](https://github.com/grpc/grpc-java/blob/master/documentation/server-reflection-tutorial.md#grpc-server-reflection-tutorial) since the 3.38.0 version.
+It's mean that users can execute calls from console or scripts via [grpcurl](https://github.com/fullstorydev/grpcurl#grpcurl) without gRPC schema (files with proto extensions described gRPC service structure)
```
jean-michel-perli-sonsin, line 251:
```suggestion
It means that the users can execute calls from the console or through scripts via [grpcurl](https://github.com/fullstorydev/grpcurl#grpcurl) without gRPC schema (files with proto extensions describes gRPC service structure) 
```

README.md
```diff
@@ -290,6 +301,10 @@ dependencies {
 
 ## Release notes
 
+### 3.38.0
+
++ gRPC router creates server supported [grpc-service-reflection](https://github.com/grpc/grpc-java/blob/master/documentation/server-reflection-tutorial.md#grpc-server-reflection-tutorial)
```
jean-michel-perli-sonsin, line 306:
```suggestion
+ gRPC router creates server support [grpc-service-reflection](https://github.com/grpc/grpc-java/blob/master/documentation/server-reflection-tutorial.md#grpc-server-reflection-tutorial)
```

jean-michel-perli-sonsin (changes requested):
Hello, please see if you agree with the changes proposed below:

**_Readme.md_**
Line 60: It **can also** be called by using `createFromArguments(args)` with **the** arguments `--namespace`, `--boxName` and `--contextName`. 

Line 82: messageRecursionLimit - an integer number denotes how deep **the** nested protobuf message might be, **set by default to** 100

Line 103: queues - the required settings defines all **the** pins for an application

Line 126: `NOT_EQUAL` - the filter passes if the field **is not** equal **to the** exact value

Line 175: enables database schema initialization if Cradle is used. By default, it **has a** value of `false`.

Line 200: **The user** needs to have authentication with **the** service account token that has the necessary access to read CRs and secrets from the specified namespace.

Line 251/262: in order for the metrics to be exported, you **will also** need to create an instance of CommonFactory.

Line 256/267: number of published message bytes to RabbitMQ. The message is **intended** for any data transferred via RabbitMQ, for example, th2 batch message or event or custom content

Line 257/268: quantity of published messages to RabbitMQ. The message is **intended** for any data transferred via RabbitMQ, for example, th2 batch message or event or custom content

Line 258/269: The message is **intended** for any data transferred via RabbitMQ, for example, th2 batch message or event or custom content

Line 259/270: **message processing time (in seconds) during subscription from RabbitMQ**. The message is **intended** for any data transferred via RabbitMQ, for example, th2 batch message or event or custom content

Line 280/291: To be able to use test extensions **,** please fill build.gradle as in **the**  example below: 

Line 319/334: Added **the** ability to read dictionaries by aliases and as group of all available aliases

Line 327/342: (if **prefetch count** is requested and no messages are acknowledged **,** the reading from the queue will be suspended).

Line 372/387: Fix **the** problem with filtering by `message_type` in MessageGroupBatch
 router

Line 449/464: Added **the** ability for resubscribe on canceled subscriber.

Line 458/473: Added the Event.toListBatchProto method to **convert an** instance of the Event class to list of event batches.

Line 471/486: The `toMetadataFilter` returns `null` **if** the original metadata has nothing to compare

Line 482/497: Added the toRootMessageFilter method to convert **a** message to root message filter

Line 499/514: NOTE: one of the methods was not restored and **an** update to this version might require manual update for your code.

**_gradle.properties_**
Line 2: Copyright **2020-2022** Exactpro (Exactpro Systems Limited)

In case of any questions, please let me know.

Thanks,

OptimumOpium (dismissed):

OptimumOpium (approved):

PR #206, opened 2022-05-27
Any.toValue() / Any.toValueFilter() > Add support for sequences and s…

ivandruzhinin:
…treams

### DIFF
--- a/src/main/kotlin/com/exactpro/th2/common/value/ValueFilterUtils.kt
+++ b/src/main/kotlin/com/exactpro/th2/common/value/ValueFilterUtils.kt
@@ -26,6 +26,7 @@ import com.exactpro.th2.common.grpc.ValueFilter
 import com.exactpro.th2.common.grpc.ValueFilter.KindCase.MESSAGE_FILTER
 import com.exactpro.th2.common.grpc.ValueFilter.KindCase.SIMPLE_FILTER
 import java.math.BigDecimal
+import java.util.stream.Stream
 
 fun equalValueFilter() = ValueFilter.newBuilder().setOperation(EQUAL).build()
 fun notEqualValueFilter() = ValueFilter.newBuilder().setOperation(NOT_EQUAL).build()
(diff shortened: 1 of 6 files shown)


src/main/kotlin/com/exactpro/th2/common/value/ValueFilterUtils.kt
```diff
@@ -56,6 +57,8 @@ fun Any.toValueFilter(): ValueFilter = when (this) {
     is ListValueFilter.Builder -> toValueFilter()
     is Iterator<*> -> toValueFilter()
     is Iterable<*> -> toValueFilter()
+    is Sequence<*> -> iterator().toValueFilter()
+    is Stream<*> -> iterator().toValueFilter()
```
Nikita-Smirnov-Exactpro, line 61: Maybe use a base approach for Stream and Sequence. I mean `.map(toValue).forEach(valueList::add)`

src/test/kotlin/com/exactpro/th2/common/message/TestMessageFilterUtils.kt
```diff
@@ -150,6 +150,17 @@ class TestMessageFilterUtils {
...
 
+    @Test
+    fun `test sequence and stream support`() {
+        val message = messageFilter().apply {
+            addField("sequence", sequenceOf(1, 2, 3, 4))
+            addField("stream", Stream.of(5, 6, 7, 8))
+        }
+
+        Assertions.assertEquals(listOf(1, 2, 3, 4), message.getField("sequence")!!.listFilter.valuesList.map { it.simpleFilter.toInt() })
+        Assertions.assertEquals(listOf(5, 6, 7, 8), message.getField("stream")!!.listFilter.valuesList.map { it.simpleFilter.toInt() })
```
Topru333, line 161: maybe is it time to implement list of primitive value util? 
Something like
when (type) {
  int -> message.getField("stream")!!.listFilter.valuesList.map { it.simpleFilter.toInt() }
}

src/main/kotlin/com/exactpro/th2/common/value/ValueFilterUtils.kt
ivandruzhinin: To do this we'll have to create `ListValue` first. So it's just better to reuse existing function which does exactly that

src/test/kotlin/com/exactpro/th2/common/message/TestMessageFilterUtils.kt
ivandruzhinin: I'm not sure that it's a very common use-case. But if it - let's do this

PR #220, opened 2022-06-27
[TH2-3789] Using cradle 3.2

LukaTchumburidze:


gradle.properties
```diff
@@ -13,7 +13,7 @@
 # limitations under the License.
 #
 
-release_version=3.39.2
+release_version=3.39.3
```
OptimumOpium, line 16: I think we should update the minor version because the minor version of the library is changed

PR #227, opened 2022-08-23
provided configuration for eventBatchDurationMillis

lumber1000:


LukaTchumburidze (changes requested):
I think this new property should be added in `README.md`'s `cradle.json` description part as well.

build.gradle
```diff
@@ -30,8 +30,8 @@ sourceCompatibility = 11
 targetCompatibility = 11
 
 ext {
-    cradleVersion       = '3.2.2'
-    junitVersion      = '5.8.2'
+    cradleVersion       = '3.1.3-TH2-3827-2909433977-SNAPSHOT'
```
OptimumOpium, line 33: Why did we downgrade the version?

lumber1000: I used this version according to task - "Update cradle version to the one from [TH2-3827]". Looks like there is no newer version which contains eventBatchDurationMillis parameter.

build.gradle
```diff
@@ -4,7 +4,8 @@ buildscript {
     }
 
     ext {
-        kotlin_version = "1.5.32"
+        kotlin_version = "1.6.21"
+        jackson_version = "2.13.4"
```
OptimumOpium, line 8: Please, remove the version for jackson. The version goes from BOM

build.gradle
```diff
@@ -4,7 +4,8 @@ buildscript {
     }
 
     ext {
-        kotlin_version = "1.5.32"
+        kotlin_version = "1.6.21"
```
OptimumOpium, line 7: Please, return back the version of Kotlin. This is not a change for updating this version

OptimumOpium (changes requested):

OptimumOpium (approved):

PR #228, opened 2022-08-24
[TH2-4137] Update bom and dependencies to exclude vulnerabilities

OptimumOpium:


build.gradle
```diff
@@ -166,54 +168,64 @@ tasks.register('integrationTest', Test) {
 }
 
 dependencies {
-    api platform("com.exactpro.th2:bom:3.2.0")
-    api ("com.exactpro.th2:cradle-core:${cradleVersion}")
+    api platform("com.exactpro.th2:bom:4.0.0-vulnerabilities-2919290091-SNAPSHOT")
```
Nikita-Smirnov-Exactpro, line 171: Replace to release version

build.gradle
```diff
@@ -166,54 +168,64 @@ tasks.register('integrationTest', Test) {
...
-    implementation "io.grpc:grpc-services:1.32.1" //FIXME: add to bom
+    implementation "io.grpc:grpc-services"
 
     implementation "com.rabbitmq:amqp-client"
 
     implementation "org.jetbrains:annotations"
 
     implementation "org.apache.commons:commons-lang3"
-    implementation "org.apache.commons:commons-collections4:4.4"
+    api("org.apache.commons:commons-collections4:4.4") {
```
Nikita-Smirnov-Exactpro, line 194: Maybe add it to bom too

build.gradle
```diff
@@ -166,54 +168,64 @@ tasks.register('integrationTest', Test) {
...
+        because('multimap in MqRouterFilterConfiguration')
+    }
     implementation "org.apache.commons:commons-text:1.9"
     implementation "commons-cli:commons-cli:1.5.0"
```
Nikita-Smirnov-Exactpro, line 198: Maybe add it to bom too

build.gradle
```diff
@@ -166,54 +168,64 @@ tasks.register('integrationTest', Test) {
...
     implementation "com.fasterxml.jackson.dataformat:jackson-dataformat-yaml"
-    // TODO: need to update jackson kotlin module to get rid of kotlin-reflect 1.3.72
     implementation "com.fasterxml.jackson.module:jackson-module-kotlin"
 
-    implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk8:${kotlin_version}"
+    implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk8"
 
     implementation 'com.fasterxml.uuid:java-uuid-generator:4.0.1'
 
     implementation 'io.github.microutils:kotlin-logging:2.1.21'
```
OptimumOpium, line 215: It is commonly used dependency. Probably, need to be moved to bom or declared as `api` dependecy

build.gradle
```diff
@@ -166,54 +168,64 @@ tasks.register('integrationTest', Test) {
...
 
     implementation 'io.prometheus:simpleclient'
     implementation 'io.prometheus:simpleclient_hotspot'
     implementation 'io.prometheus:simpleclient_httpserver'
     implementation 'io.prometheus:simpleclient_log4j2'
-    implementation 'io.prometheus:simpleclient_log4j'
 
-    implementation 'io.fabric8:kubernetes-client:4.13.0'
+    // fix vulnerability in transitive dependency
+    implementation 'com.squareup.okhttp3:okhttp:4.10.0'
```
OptimumOpium, line 227: Where it goes from? Please, add the reason why this is here in `bacause` block
```suggestion
    implementation ('com.squareup.okhttp3:okhttp:4.10.0') {
      because('fix vulnerability in transitive dependency for <original dependency name>')
    }
```

Xanclry:
I also checked the work of the new version of the kubernates client in a cluster, everything works fine.

build.gradle
OptimumOpium: If we have added it into the bom you can remove version from here. The same for other dependencies which you have added to the bom

build.gradle
```diff
@@ -166,54 +168,65 @@ tasks.register('integrationTest', Test) {
...
+    api("com.fasterxml.jackson.core:jackson-annotations") {
+        because('providee ability to use jackson annotations in components')
+    }
     implementation "com.fasterxml.jackson.datatype:jackson-datatype-jsr310"
     implementation "com.fasterxml.jackson.dataformat:jackson-dataformat-yaml"
-    // TODO: need to update jackson kotlin module to get rid of kotlin-reflect 1.3.72
     implementation "com.fasterxml.jackson.module:jackson-module-kotlin"
 
-    implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk8:${kotlin_version}"
+    implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk8"
```
OptimumOpium, line 211: I am not sure but probably you don't need it here

Nikita-Smirnov-Exactpro (approved):

PR #229, opened 2022-09-01
[TH2-4170] Correct filters behavior

OptimumOpium:


src/main/kotlin/com/exactpro/th2/common/schema/message/impl/rabbitmq/group/util/GroupUtil.kt
```diff
@@ -0,0 +1,51 @@
...
+import com.exactpro.th2.common.grpc.MessageGroupBatch
+
+/**
+ * Transfers messages that match the [filterFunction] from current group to new group in [batchBuilder].
+ *
+ * If all messages are filtered no group will be added.
+ *
+ * If all messages match the filter the current group will be added into [batchBuilder] as is
+ */
+internal fun MessageGroup.filterMessagesToNewGroup(
```
ivandruzhinin, line 29: Maybe make it inline?

src/main/kotlin/com/exactpro/th2/common/schema/message/impl/rabbitmq/group/util/GroupUtil.kt
```diff
@@ -0,0 +1,51 @@
...
+    batchBuilder: MessageGroupBatch.Builder,
+    filterFunction: (AnyMessage) -> Boolean,
+): List<AnyMessage> {
+    val matchCount = messagesList.count { filterFunction(it) }
```
ivandruzhinin, line 33: If filter-function is computation heavy invoking it twice to avoid new group creation might result in a negative performance yield. Not sure how lightweight our filters are

ivandruzhinin (approved):

src/main/kotlin/com/exactpro/th2/common/schema/message/impl/rabbitmq/group/RabbitMessageGroupBatchSubscriber.kt
```diff
@@ -57,25 +58,24 @@ class RabbitMessageGroupBatchSubscriber(
...
-                            MESSAGE_GROUP_DROPPED_SUBSCRIBE_TOTAL
-                        )
-                    }
+        val newBatch = MessageGroupBatch.newBuilder()
+
+        batch.groupsList.forEach { group ->
+            val dropped = group.filterMessagesToNewGroup(newBatch) { callFilterFunction(it, filters) }
+            if (dropped.isNotEmpty()) {
+                logger.debug {
+                    "Skipped ${dropped.size} message(s) from group because it(they) didn't match any filters: ${dropped.joinToString { it.toJson() }}"
```
Nikita-Smirnov-Exactpro, line 67: Message can be too large. Please print short version of it, for example metadata or message id.
NOTE: As I know users set unique sequence to message id in scripts to trace messages

src/main/kotlin/com/exactpro/th2/common/schema/message/impl/rabbitmq/group/util/GroupUtil.kt
Nikita-Smirnov-Exactpro: Now filters are lightweight but they will be heavy after adding business filter and switching to Cbor body

src/main/kotlin/com/exactpro/th2/common/schema/message/impl/rabbitmq/group/util/GroupUtil.kt
OptimumOpium: The filter function should not be able to stop this function work. So, I think it is better to keep it as is

src/main/kotlin/com/exactpro/th2/common/schema/message/impl/rabbitmq/group/RabbitMessageGroupBatchSubscriber.kt
OptimumOpium: Changed to `logId`

src/main/kotlin/com/exactpro/th2/common/schema/message/impl/rabbitmq/group/util/GroupUtil.kt
ivandruzhinin: You can add `crossinline` modifier to it to prohibit non-local returns

src/main/kotlin/com/exactpro/th2/common/schema/message/impl/rabbitmq/group/util/GroupUtil.kt
```diff
@@ -0,0 +1,48 @@
...
+    if (messagesCount == matched.size) {
+        batchBuilder.addGroups(this)
+        return emptyList()
+    }
+    if (matched.isEmpty()) {
+        return messagesList
+    }
+    val newGroup = batchBuilder.addGroupsBuilder()
+    matched.forEach { newGroup.addMessages(it.message) }
+    return dropped.map { it.message }
```
ivandruzhinin, line 47:
```suggestion
    val (matched, dropped) = messagesList.partition { filterFunction(it) }

    when {
        dropped.isEmpty() -> batchBuilder.addGroups(this)
        else -> batchBuilder.addGroupsBuilder().addAllMessages(matched)
    }

    return dropped
```

ivandruzhinin (changes requested):

src/main/kotlin/com/exactpro/th2/common/schema/message/impl/rabbitmq/group/util/GroupUtil.kt
OptimumOpium: Thank you, my bad. I don't know why I have added the extra wrapping instead of simple partitioning by filter.
But I want to keep all created lists local if it is possible so I will change it a bit

src/main/kotlin/com/exactpro/th2/common/schema/message/impl/rabbitmq/group/util/GroupUtil.kt
OptimumOpium: Indeed. Forgot about that, thank you

src/main/kotlin/com/exactpro/th2/common/schema/message/impl/rabbitmq/group/util/GroupUtil.kt
ivandruzhinin: > I want to keep all created lists local

Why? If you fear that `dropped` list could be modified from Java you can wrap it in `Collections.unmodifiableList`. Otherwise I see no reason in doing so.

src/main/kotlin/com/exactpro/th2/common/schema/message/impl/rabbitmq/group/util/GroupUtil.kt
Nikita-Smirnov-Exactpro: Looks fine for me

ivandruzhinin (approved):

PR #230, opened 2022-09-05
Option to restrict max message size

isengrims:


src/main/java/com/exactpro/th2/common/schema/grpc/router/impl/DefaultStubStorage.java
```diff
@@ -122,6 +122,7 @@ public T getStub(@NotNull Message message, @NotNull AbstractStub.StubFactory<T>
                             .keepAliveTime(clientConfiguration.getKeepAliveInterval(), TimeUnit.SECONDS)
                             .usePlaintext()
                             .intercept(new GrpcInterceptor(service.pinName, methodInvokeCounter, requestBytesCounter, responseBytesCounter))
+                            .maxInboundMessageSize(endpoint.getMaxMessageSize())
```
Nikita-Smirnov-Exactpro, line 125: You forget about AbstractGrpcRouter.getOrCreateChannel method.

Nikita-Smirnov-Exactpro: https://github.com/th2-net/th2-common-j/pull/181/files

Nikita-Smirnov-Exactpro: Also, please have a look at the service side in the AbstractGrpcRouter.startServer method. The NettyServerBuilder class has the same maxInboundMessageSize method

src/main/kotlin/com/exactpro/th2/common/schema/grpc/configuration/GrpcConfiguration.kt
```diff
@@ -46,7 +47,8 @@ data class GrpcClientConfiguration(
 data class GrpcEndpointConfiguration(
     @JsonProperty(required = true) var host: String,
     @JsonProperty(required = true) var port: Int = 8080,
-    var attributes: List<String?> = emptyList()
+    @JsonProperty var maxMessageSize: Int = GrpcUtil.DEFAULT_MAX_MESSAGE_SIZE,
```
Nikita-Smirnov-Exactpro, line 50: maxMessageSize should be use for client and server, also, users use only default value for this option I think.
Let's move this option from EndpointConfiguration to the GrpcConfiguration class

Nikita-Smirnov-Exactpro (changes requested):

Nikita-Smirnov-Exactpro (approved):

PR #235, opened 2022-10-18
[TH2-4299] Manual Acknowledgement refinement in common library

Xanclry:


src/main/kotlin/com/exactpro/th2/common/schema/message/ManualAckDeliveryCallback.kt
```diff
@@ -26,11 +26,22 @@ fun interface ManualAckDeliveryCallback {
...
+    @Deprecated(
+        "This method does not provide all necessary information about a message",
+        ReplaceWith("handle(deliveryMetadata, delivery, confirmProcessed)")
+    )
     @Throws(IOException::class)
     fun handle(consumerTag: String, delivery: Delivery, confirmProcessed: Confirmation)
 
-    fun interface Confirmation {
+    @Throws(IOException::class)
+    fun handle(deliveryMetadata: DeliveryMetadata, delivery: Delivery, confirmProcessed: Confirmation) {
```
Nikita-Smirnov-Exactpro, line 37: We should move this change to the ConfirmationMessageListener interface because it is part of public common API.
the handle method of this class should be extended via DeliveryMetadata

Nikita-Smirnov-Exactpro (changes requested):

PR #258, opened 2023-02-27
Vulnerabilities book and page

Nikita-Smirnov-Exactpro:


README.md
```diff
@@ -1,4 +1,4 @@
-# th2 common library (Java) (5.1.1)
+# th2 common library (Java) (5.2.0)
```
OptimumOpium, line 1: Do we really need to update version if we don't even have a first release?

OptimumOpium (approved):

README.md
Nikita-Smirnov-Exactpro: I think yes, because we made versioning document to follow changes in projects for development.

PR #266, opened 2023-05-16
Json raw body

OptimumOpium:


src/main/kotlin/com/exactpro/th2/common/schema/message/impl/rabbitmq/transport/Message.kt
```diff
@@ -25,7 +25,7 @@ interface Message<T> : Cleanable {
     /** The metadata is not mutable by default */
     var metadata: MutableMap<String, String>
     var protocol: String
-    var body: T
+    val body: T
```
Nikita-Smirnov-Exactpro, line 28: Why we close set option? It looks strange on the other mutable fields background

OptimumOpium: You're right, probably, we should return the mutability here...
At least until we migrate to immutable objects (if we do)

Nikita-Smirnov-Exactpro (approved):

Nikita-Smirnov-Exactpro (approved):

src/test/kotlin/com/exactpro/th2/common/schema/message/impl/rabbitmq/transport/TransportUtilsTest.kt
```diff
@@ -165,20 +178,75 @@ class TransportUtilsTest {
...
+            }
+            setType("test-type")
+        }.build()
+
+        val proto = transport.toProto(BOOK_NAME, SESSION_GROUP)
+        assertEquals(BOOK_NAME, proto.metadata.id.bookName)
+        assertEquals(SESSION_GROUP, proto.metadata.id.connectionId.sessionGroup)
+        proto.toTransport().apply {
+            assertEquals(transport, this)
+        }
```
OptimumOpium, line 245: Honestly, it is not the right way to test the conversion, I think... If you want to test conversion you need separate tests for `transport -> proto` and `proto -> transport` and do not join them.

src/main/kotlin/com/exactpro/th2/common/message/MessageUtils.kt
```diff
@@ -457,14 +492,16 @@ fun getDebugString(className: String, ids: List<MessageID>): String {
...
+    JsonFormat.printer().includingDefaultValueFields().let {
+        (if (short) it.omittingInsignificantWhitespace() else it).print(this)
+    }
 
-fun <T: com.google.protobuf.Message.Builder> T.fromJson(json: String) : T = apply {
+fun <T : com.google.protobuf.Message.Builder> T.fromJson(json: String): T = apply {
     JsonFormat.parser().ignoringUnknownFields().merge(json, this)
 }
 
+@Deprecated("Moved to common-utils-j/com.exactpro.th2.common.utils.message.MessageUtils")
```
OptimumOpium, line 504: It is kotlin, so we can add a replacement suggestion to the deprecation annotation. It does not check the presents of the class but will it automates the migration

src/main/kotlin/com/exactpro/th2/common/schema/message/impl/rabbitmq/transport/ParsedMessage.kt
```diff
@@ -102,7 +99,6 @@ class ParsedMessage private constructor(
     interface FromMapBuilder : Builder<FromMapBuilder> {
         fun setBody(body: Map<String, Any?>): FromMapBuilder
         fun bodyBuilder(): MapBuilder<String, Any?>
-        fun addField(name: String, value: Any?): FromMapBuilder
```
OptimumOpium: Why have you decided to remove that method? It allows adding a field without breaking the methods chain

src/main/kotlin/com/exactpro/th2/common/schema/message/impl/rabbitmq/transport/ParsedMessage.kt
```diff
@@ -90,9 +89,7 @@ class ParsedMessage private constructor(
         val type: String
 
         fun setType(type: String): T
-        fun metadataBuilder(): ImmutableMap.Builder<String, String>
-        fun addMetadataProperty(key: String, value: String): T
```
OptimumOpium: The same here - why you decided to remove that method? It looks very helpful when we use the builder from Java  where no `apply` function exists

src/main/kotlin/com/exactpro/th2/common/schema/message/impl/rabbitmq/transport/ParsedMessage.kt
Nikita-Smirnov-Exactpro: This method can confuse users when they use `.apply {` approach to fill message. I have add a separate extension method for that if you want work with builder directly.

OptimumOpium: They can't use the `apply` method in Java... The extension method you added can be used conveniently only from Kotlin. In Java it will become a static method

src/test/kotlin/com/exactpro/th2/common/schema/message/impl/rabbitmq/transport/TransportUtilsTest.kt
Nikita-Smirnov-Exactpro: Implemented in the common-utils project

src/main/kotlin/com/exactpro/th2/common/schema/message/impl/rabbitmq/transport/ParsedMessage.kt
Nikita-Smirnov-Exactpro: Reverted

OptimumOpium (commented):
Looks good to me

src/main/kotlin/com/exactpro/th2/common/schema/message/impl/rabbitmq/transport/ParsedMessage.kt
```diff
@@ -161,14 +161,18 @@ class ParsedMessage private constructor(
     }
 }
 
+@Suppress("PropertyName")
 private sealed class BaseParsedBuilder<out T : ParsedMessage.Builder<T>> : ParsedMessage.Builder<T> {
     protected var idBuilder: MessageId.Builder? = null
     protected var id: MessageId? = MessageId.DEFAULT
-    protected var eventId: EventId? = null
+    protected var _eventId: EventId? = null
```
OptimumOpium, line 168: You can just declare it as var here and make the setter private or protected
```suggestion
    override var eventId: EventId? = null
      private set
```
Should do the work

OptimumOpium (commented):
Looks fine for me

src/main/kotlin/com/exactpro/th2/common/schema/message/impl/rabbitmq/transport/ParsedMessage.kt
```diff
@@ -144,6 +144,8 @@ class ParsedMessage private constructor(
 
 
     companion object {
+        val EMPTY = ParsedMessage(type = "", body = emptyMap())
```
OptimumOpium, line 147: Why do you need an empty value for a message?

Nikita-Smirnov-Exactpro: Yes, conn-http expects pair of message: Parsed and Raw. the conn uses empty message when one of them don't come.

OptimumOpium: Then maybe it should use `null` instead of an EMTPY value... The null is a great indicator that we don't have a value

Nikita-Smirnov-Exactpro: conn-http works with messages as they have empty body, metadata and so on

build.gradle
```diff
@@ -180,6 +181,7 @@ dependencies {
     api 'com.exactpro.th2:grpc-common:4.2.0-dev'
     api "com.exactpro.th2:cradle-core:${cradleVersion}"
     api 'io.github.microutils:kotlin-logging:2.1.21'
+    api 'io.netty:netty-buffer'
```
OptimumOpium, line 184: I think we should add a `because` description here. Because this is the first candidate to move into th2-protocol lib when we separate it from common lib

OptimumOpium (commented):
Looks good to me

src/main/kotlin/com/exactpro/th2/common/message/MessageUtils.kt
```diff
@@ -517,7 +517,7 @@ fun Message.toTreeTable(): TreeTable = TreeTableBuilder().apply {
 val MessageIDOrBuilder.isValid: Boolean
     get() = bookName.isNotBlank()
             && hasConnectionId() && connectionId.sessionAlias.isNotBlank()
-            && hasTimestamp() && timestamp.seconds > 0 && timestamp.nanos > 0
+            && hasTimestamp() && timestamp.seconds > 0
```
OptimumOpium, line 520: We must consider the case when seconds = 0 but nanos is not
```suggestion
            && hasTimestamp() && (timestamp.seconds > 0 || timestamp.nanos > 0)
```

Nikita-Smirnov-Exactpro:
@OptimumOpium let's merge this change. I'm going to solve vulnerability problem in the separate-executor method

OptimumOpium:
> @OptimumOpium let's merge this change. I'm going to solve vulnerability problem in the separate-executor method

Okay, not objections from my side

PR #268, opened 2023-06-06
Update cradle version

OptimumOpium:
The old dev version has incompatible changes in API that causes runtime errors when we try to use common 5.2.0-dev with any new Cradle version.

Here we update the cradle version to match the new API


build.gradle
```diff
@@ -29,7 +29,7 @@ group = 'com.exactpro.th2'
 version = release_version
 
 ext {
-    cradleVersion       = '5.0.0-dev-version-5-2567311063-SNAPSHOT'
+    cradleVersion       = '5.0.2-dev-version-5-5176643613-bf4794b-SNAPSHOT'
```
Nikita-Smirnov-Exactpro, line 32: Please use cradle dev release

OptimumOpium: There is not dev release for cradle

OptimumOpium: @Nikita-Smirnov-Exactpro If you insist I will add the corresponding workflow to Cradle repository

Nikita-Smirnov-Exactpro: Could make it to clarify what version do we use?

OptimumOpium: We already have the exact commit that corresponds to the version.
But yes, I will add the dev release workflow to Cradle

PR #271, opened 2023-07-21
Book, session group and protocol message filtering added.

lumber1000:


lumber1000:
https://github.com/th2-net/th2-common-j/pull/272

PR #275, opened 2023-09-04
[TH2-5007] MessageSubscriber supports single listener

Nikita-Smirnov-Exactpro:
* The unsubscribe method of SubscribtionMonitor interface cancels RabbitMQ subscribtion


Nikita-Smirnov-Exactpro:
> Also, I believe that we should also have an integration test for that issue:
> 
> 1. Handle delivery with manual ack
> 2. Unsubscribe using `Subscription`
> 3. Send one more message
> 4. Confirm the previous message
> 5. Subscribe again
> 6. Expect to receive that message from point 3

implemented

src/main/kotlin/com/exactpro/th2/common/schema/message/DeliveryMetadata.kt
```diff
@@ -17,5 +17,6 @@ package com.exactpro.th2.common.schema.message
 
 data class DeliveryMetadata(
     val consumerTag: String,
+    val deliveryTag: Long,
```
OptimumCode, line 20: Why do we need a delivery tag here? It is a rabbitmq specific but we expose it as a public API... The only place you use it - tests. Probably, the content itself should be enough for tests...

src/test/kotlin/com/exactpro/th2/common/schema/message/impl/rabbitmq/AbstractRabbitRouterIntegrationTest.kt
```diff
@@ -0,0 +1,193 @@
...
+                val monitor = router.subscribeWithManualAck({ deliveryMetadata, message, confirmation ->
+                    queue.put(Delivery(message, deliveryMetadata.deliveryTag, deliveryMetadata.isRedelivered, confirmation))
+                })
+
+                try {
+                    // RabbitMQ doesn't resend messages after resubscribe using the same connection and channel
+                    assertNull(queue.poll(1, TimeUnit.SECONDS))
+                } finally {
+                    monitor.unsubscribe()
+                }
```
OptimumCode, line 128: Do we need to test rabbitmq here?.. This part does exactly this in my opinion.
We just need to make sure that after resubscribing we receive the expected messages from the queue, I think

src/test/kotlin/com/exactpro/th2/common/schema/message/impl/rabbitmq/TestRouter.kt
```diff
@@ -0,0 +1,52 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.exactpro.th2.common.schema.message.impl.rabbitmq
+
+import com.exactpro.th2.common.schema.message.ConfirmationListener
+import com.exactpro.th2.common.schema.message.MessageSender
+import com.exactpro.th2.common.schema.message.MessageSubscriber
+
+internal class TestRouter : AbstractRabbitRouter<String>() {
```
OptimumCode, line 22: I think we could use the existing custom router for that purpose

src/test/kotlin/com/exactpro/th2/common/schema/message/impl/rabbitmq/AbstractRabbitRouterIntegrationTest.kt
Nikita-Smirnov-Exactpro: I spend some times to understand that rabbitmq doesn't resend message when we use the same connection and channel. I would like to illustrate this case in the test

OptimumCode: Then maybe we should add this information to javadoc for subscribe methods, what do you think? Because I don't believe that anyone who uses the common lib will go to its tests to check how the method behaves in this particular case

OptimumCode (approved):

src/main/java/com/exactpro/th2/common/schema/message/MessageRouter.java
```diff
@@ -61,7 +61,12 @@ default void init(@NotNull MessageRouterContext context, @NotNull MessageRouter<
     ExclusiveSubscriberMonitor subscribeExclusive(MessageListener<T> callback);
 
     /**
-     * Listen <b>ONE</b> RabbitMQ queue by intersection schemas queues attributes
+     * Listen <b>ONE</b> RabbitMQ queue by intersection schemas queues attributes.
+     * Restrictions:
+     * You can create only one subscription to th2 pin using any subscribe* functions. This method throws RuntimeException if the th2 pin is matched by passed attributes already has active subscription.
```
OptimumCode, line 66: I think information about throwing an exception should be moved to a corresponding block `@throws`

OptimumCode (approved):

PR #277, opened 2023-09-18
[TH2-5069] Provided the ability to define configs directory using the…

Nikita-Smirnov-Exactpro:
… `th2.common.configuration-directory` environment variable


src/test/kotlin/com/exactpro/th2/common/schema/message/impl/rabbitmq/transport/CodecsTest.kt
```diff
@@ -154,6 +154,7 @@ class CodecsTest {
 
                 assertEquals(parsedMessage, decoded, "unexpected parsed result decoded")
                 assertEquals(
+                    //FIXME: org.opentest4j.AssertionFailedError: unexpected raw body ==> expected: <{"field":"16:36:38.035420"}> but was: <{"field":"16:36:38.03542"}>
```
OptimumCode, line 157: We should add a corresponding formatter to each type:

LocalDate: value.toString()
LocalTime: value.format(DateTimeFormatter.ISO_LOCAL_TIME)
LocalDateTime: value.format(DateTimeFormatter.ISO_LOCAL_DATE_TIME)
Instant: value.toString()

src/main/java/com/exactpro/th2/common/schema/factory/CommonFactory.java
```diff
@@ -632,6 +550,43 @@ public InputStream readDictionary(DictionaryType dictionaryType) {
...
+    }
+
+    static @NotNull Path getConfigPath(@Nullable String cmdPath) {
+        String pathString = StringUtils.trim(cmdPath);
+        if (pathString != null) {
+            Path path = Paths.get(pathString);
+            if (Files.exists(path) && Files.isDirectory(path)) {
+                return path;
+            } else {
+                LOGGER.warn("'" + cmdPath + "' config directory passed via CMD doesn't exist");
```
OptimumCode, line 581:
```suggestion
                LOGGER.warn("'" + cmdPath + "' config directory passed via CMD doesn't exist or it is not a directory");
```

src/main/java/com/exactpro/th2/common/schema/factory/CommonFactory.java
```diff
@@ -632,6 +550,43 @@ public InputStream readDictionary(DictionaryType dictionaryType) {
...
+        return CONFIG_DEFAULT_PATH;
+    }
+
+    static @NotNull Path getConfigPath(@Nullable String cmdPath) {
+        String pathString = StringUtils.trim(cmdPath);
+        if (pathString != null) {
+            Path path = Paths.get(pathString);
+            if (Files.exists(path) && Files.isDirectory(path)) {
+                return path;
+            } else {
```
OptimumCode, line 580: No need to `else` brach here because `if` branch calls `return`

src/main/java/com/exactpro/th2/common/schema/factory/CommonFactory.java
```diff
@@ -632,6 +550,43 @@ public InputStream readDictionary(DictionaryType dictionaryType) {
...
     }
 
+    static @NotNull Path getConfigPath() {
+        String pathString = System.getenv(TH2_COMMON_CONFIGURATION_DIRECTORY_ENVIRONMENT_VARIABLE);
+        if (pathString != null) {
+            Path path = Paths.get(pathString);
+            if (Files.exists(path) && Files.isDirectory(path)) {
+                return path;
+            } else {
+                LOGGER.warn("'{}' config directory passed via '{}' environment variable doesn't exist",
```
OptimumCode, line 560:
```suggestion
                LOGGER.warn("'{}' config directory passed via '{}' environment variable doesn't exist or it is not a directory",
```

src/main/java/com/exactpro/th2/common/schema/factory/CommonFactory.java
```diff
@@ -632,6 +550,43 @@ public InputStream readDictionary(DictionaryType dictionaryType) {
         }
     }
 
+    static @NotNull Path getConfigPath() {
+        String pathString = System.getenv(TH2_COMMON_CONFIGURATION_DIRECTORY_ENVIRONMENT_VARIABLE);
+        if (pathString != null) {
+            Path path = Paths.get(pathString);
+            if (Files.exists(path) && Files.isDirectory(path)) {
+                return path;
+            } else {
```
OptimumCode, line 559: Also, no need for `else` branch. `If` branch will exit the function if executed

src/test/kotlin/com/exactpro/th2/common/schema/message/impl/rabbitmq/transport/CodecsTest.kt
Nikita-Smirnov-Exactpro: Do you have example or link to solution?

OptimumCode: No, I don't have an example or link to the solution. I just checked what is used by Jackson module to serialize those classes and how `toString` method is implemented in those classes

OptimumCode: You could do something like this:
```kotlin
val testData = listOf<Pair<TemporalAccessor, (TemporalAccessor) -> String>>(
  LocalDate.now() to  value::toString,
  LocalTime.now() to  DateTimeFormatter.ISO_LOCAL_TIME::format,
  LocalDateTime.now() to DateTimeFormatter.ISO_LOCAL_DATE_TIME::format,
  Instant.now() to value::toString,
)

File: src/test/kotlin/com/exactpro/th2/common/schema/message/impl/rabbitmq/transport/CodecsTest.kt [hunk 6]
```diff
@@ -129,11 +131,14 @@ class CodecsTest {
...
+        LocalTime.parse("16:36:38.035420").toString()
+        val testData = listOf<Pair<TemporalAccessor, (TemporalAccessor) -> String>>(
+            LocalDate.now() to TemporalAccessor::toString,
+            LocalTime.now() to DateTimeFormatter.ISO_LOCAL_TIME::format,
+            // Check case when LocalTime.toString() around nanos to 1000
+            LocalTime.parse("16:36:38.035420") to DateTimeFormatter.ISO_LOCAL_TIME::format,
+            LocalDateTime.now() to DateTimeFormatter.ISO_LOCAL_DATE_TIME::format,
+            Instant.now() to TemporalAccessor::toString,
         )
         return testData.map {
```
OptimumCode, line 143: You can use destruction declaration in `map` lambda for more readable code
```suggestion
        return testData.map { (value, formatter) ->
```

OptimumCode (approved):

OptimumCode (approved):

PR #278, opened 2023-10-06
Added microservice entry-point

Nikita-Smirnov-Exactpro:


src/main/kotlin/com/exactpro/th2/common/microservice/ApplicationContext.kt
```diff
@@ -0,0 +1,28 @@
...
+
+import com.exactpro.th2.common.schema.factory.AbstractCommonFactory
+
+/**
+ * @param registerResource: register all resources which must be closed in reverse registration order.
+ * @param onPanic: call this method when application can not correct work anymore.
+ */
+class ApplicationContext(
+    val commonFactory: AbstractCommonFactory,
+    val registerResource: (AutoCloseable) -> Unit,
```
OptimumCode, line 26: I think it would be convenient to provide a name for the resource. It will make logging messages more informative. I also think that we should use `BiConsumer<T, U>` here (it will be more convenient for Java)
```suggestion
    val registerResource: BiConsumer<String, AutoCloseable>,
```

src/main/kotlin/com/exactpro/th2/common/microservice/ApplicationContext.kt
```diff
@@ -0,0 +1,28 @@
...
+    val onPanic: (Throwable?) -> Unit,
```
OptimumCode, line 27: Here I would also suggest using `Consumer` interface instead of Kotlin lambda

src/main/kotlin/com/exactpro/th2/common/microservice/Main.kt
```diff
@@ -0,0 +1,151 @@
...
+                                "and $APPLICATION_FACTORY_SYSTEM_PROPERTY system property isn't specified," +
+                                "instances: $instances"
+                    )
+                }
+            }
+        }
+    }
+}
+
+fun main(args: Array<String>) {
```
OptimumCode, line 107: From my own experience, having a main method in another project is not really convenient for understanding what is going on when you run the application...
If we take a look at Spring Boot (https://spring.io/quickstart - step 3) the main method is defined in the application code (not in the library). The library only provides users with a starter method:
```java
SpringApplication.run(DemoApplication.class, args);
```

Probably, we should go with the same approach, in my opinion.
We can still use ServiceLoader mechanism under the hood but move the main method from library to the application code

src/main/kotlin/com/exactpro/th2/common/schema/configuration/IConfigurationProvider.kt
```diff
@@ -0,0 +1,20 @@
...
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+package com.exactpro.th2.common.schema.configuration
+
+interface IConfigurationProvider {
+    fun <T : Any> load(configClass: Class<T>, alias: String, default: () -> T): T
```
OptimumCode, line 19: What if I don't have a default option?

We should have a method that can load required configurations

src/main/kotlin/com/exactpro/th2/common/schema/configuration/impl/JsonConfigurationProvider.kt
```diff
@@ -0,0 +1,90 @@
...
+            val content = SUBSTITUTOR.get().replace(sourceContent)
+            requireNotNull(MAPPER.readValue(content, configClass)) {
+                "Parsed format of the '$alias' config alias content can't be null"
+            }
+        }.also { value ->
+            check(configClass.isInstance(value)) {
+                "Stored configuration instance of $alias config alias mismatches, " +
+                        "expected: ${configClass.canonicalName}, actual: ${value::class.java.canonicalName}"
+            }
+        } as T
```
OptimumCode, line 64: You can use `case` method from `Class` here to get rid of `UNCHECKED_CAST`
```suggestion
        }.let(configClass::cast)
```

src/main/kotlin/com/exactpro/th2/common/schema/configuration/impl/JsonConfigurationProvider.kt
```diff
@@ -0,0 +1,90 @@
...
+
+        @JvmField
+        val MAPPER = ObjectMapper().apply {
+            registerModules(
+                KotlinModule.Builder()
+                    .withReflectionCacheSize(512)
+                    .configure(KotlinFeature.NullToEmptyCollection, false)
+                    .configure(KotlinFeature.NullToEmptyMap, false)
+                    .configure(KotlinFeature.NullIsSameAsDefault, false)
+                    .configure(KotlinFeature.SingletonSupport, false)
```
OptimumCode, line 82: I think we should enable SingletonSupport. Because if it is disabled new instances of kotlin objects might be created

src/main/kotlin/com/exactpro/th2/common/microservice/IApplicationFactory.kt
```diff
@@ -0,0 +1,162 @@
...
+     * Close resources shared between created application.
+     */
+    override fun close() {}
+
+    companion object {
+        private val K_LOGGER = KotlinLogging.logger {}
+        private const val APPLICATION_FACTORY_SYSTEM_PROPERTY = "th2.microservice.application-factory"
+        private const val MONITOR_NAME = "microservice_main"
+        @JvmStatic
+        fun run(factory: IApplicationFactory?, vararg args: String) {
```
OptimumCode, line 51: We can set the `null` default value to `factory` and add `@JvmOverloads` annotations to have a method that accepts only varargs

src/main/java/com/exactpro/th2/common/schema/factory/CommonFactory.java
```diff
@@ -159,6 +155,11 @@ protected Path getOldPathToDictionariesDir() {
     protected ConfigurationManager getConfigurationManager() {
         return configurationManager;
     }
+
+    IConfigurationProvider getConfigurationProvider() {
+        return configurationProvider;
+    }
```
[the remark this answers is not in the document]
OptimumCode: After discussion, the method should be left but made public

src/main/kotlin/com/exactpro/th2/common/schema/configuration/impl/JsonConfigurationProvider.kt
```diff
@@ -0,0 +1,114 @@
...
+import java.io.InputStream
+import java.nio.file.Files
+import java.nio.file.Path
+import java.util.concurrent.ConcurrentHashMap
+import java.util.function.Function
+import java.util.function.Supplier
+import kotlin.io.path.exists
+
+
+class JsonConfigurationProvider @JvmOverloads constructor(
```
OptimumCode, line 35: We probably should make the constructor private and create factory methods creating an instance of `JsonConfigurationProvider`

Nikita-Smirnov-Exactpro:
Discussion results:
Common factory constructors:
* migrate from constructors to separate static methods
* user can be able to create common factory via `craterFromArgs(vararg args: String)` and `createFromProvider(provider: IConfigurationProvider)`. Other ways should be deprecated

Common factory methods:
* `public <T> T getConfiguration(Path configPath, Class<T> configClass, ObjectMapper customObjectMapper)` should be removed because this method can be implemented out side from common factory
* `protected abstract Path getPathToCustomConfiguration()` should be removed after `createFromProvider` implementation

JsonConfigurationProvider:
* public constructor should be replaced to a static method
* class should be public because it is useful for tests

PR #279, opened 2023-10-13
Add methods to check if the field is set in builders

OptimumCode:


src/main/kotlin/com/exactpro/th2/common/schema/message/impl/rabbitmq/transport/MessageId.kt
```diff
@@ -75,13 +78,121 @@ data class MessageId(
...
         @JvmStatic
-        fun builder(): Builder = AutoBuilder_MessageId_Builder()
+        fun builder(): Builder = MessageIdBuilderImpl()
     }
-}
\ No newline at end of file
+}
+
+
+private const val SEQUENCE_NOT_SET = -1L
```
Nikita-Smirnov-Exactpro, line 93: Why we add this constant?

src/main/kotlin/com/exactpro/th2/common/schema/message/impl/rabbitmq/transport/ParsedMessage.kt
```diff
@@ -87,6 +87,7 @@ class ParsedMessage private constructor(
 
     interface Builder<out T : Builder<T>> : Message.Builder<T> {
         val type: String
+        val isTypeSet: Boolean
```
Nikita-Smirnov-Exactpro, line 90: Should we add similar properties for Parsed builder implementation

Nikita-Smirnov-Exactpro (approved):

src/main/kotlin/com/exactpro/th2/common/schema/message/impl/rabbitmq/transport/ParsedMessage.kt
OptimumCode: All parsed builders extend this interface. So, they have it already

src/main/kotlin/com/exactpro/th2/common/schema/message/impl/rabbitmq/transport/Message.kt
```diff
@@ -30,6 +30,7 @@ interface Message<T> {
 
     interface Builder<out T : Builder<T>> {
         val protocol: String
+        fun isProtocolSet(): Boolean
         val eventId: EventId?
 
         fun setId(id: MessageId): T
```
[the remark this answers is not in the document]
OptimumCode: I have added these methods only for existing getters. If we need new getters let's add them

src/main/kotlin/com/exactpro/th2/common/schema/message/impl/rabbitmq/transport/MessageId.kt
OptimumCode: Because I want to have `long` in the builder class for sequence instead of `Long`. So, we need to have an indicator that sequence is not set. The `0` looks like a valid value so I used `-1` as a marker

Nikita-Smirnov-Exactpro: Maybe use minimal long instead. 
I could be wrong, but I saw negative number in the sequence field in an act logs in a client project solution. I think that users can use this approach to trace messages before sending.

Nikita-Smirnov-Exactpro (approved):

PR #282, opened 2023-10-18
[Th2 5107] User can't change gRPC retry configuration via schema

Nikita-Smirnov-Exactpro:


README.md
```diff
@@ -491,6 +501,15 @@ dependencies {
 
 ## Release notes
 
+### 5.7.0-dev
+
+#### Fix:
++ gRPC `retryConfiguration` has been moved from grpc.json to grpc_router.json
++ the whole default gRPC retry interval is about 1 hour
```
OptimumCode, line 508: I believe this is too much for the retry... The suggest we make it 1-5 minutes max. Otherwise, we will cause scripts to wait for a long time to report an error if a component in some kind of crash-look for example...

src/main/kotlin/com/exactpro/th2/common/schema/grpc/configuration/GrpcRetryConfiguration.kt
```diff
@@ -0,0 +1,59 @@
...
+        }
+
+        return minMethodRetriesTimeout + increment
+    }
+
+    override fun getMaxAttempts(): Int = maxAttempts
+
+    override fun retryInterruptedTransaction(): Boolean {
+        return retryInterruptedTransaction
+    }
```
OptimumCode, line 58: In Kotlin you can always simplify the one-line method like this
```suggestion
    override fun retryInterruptedTransaction(): Boolean = retryInterruptedTransaction
```

src/test/kotlin/com/exactpro/th2/common/schema/grpc/router/impl/DefaultGrpcRouterTest.kt
```diff
@@ -0,0 +1,802 @@
...
+        @Test
+        override fun `delayed server start`() {
+            val clientServerBaton = Baton("client-server")
+            val future = executor.submit<Response> {
+                clientServerBaton.give("client thread started")
+                return@submit createClientSync().singleRequestSingleResponse(Request.newBuilder().setSeq(1).build())
+            }
+
+            clientServerBaton.get("wait client thread start")
+            Thread.sleep(RETRY_TIMEOUT * 2)
```
OptimumCode, line 172: Why do we need to wait here? If you need to make sure the task is executed within a timeout you can use `Assertions.assertTimeoutPreemptively` method

src/test/kotlin/com/exactpro/th2/common/schema/grpc/router/impl/DefaultGrpcRouterTest.kt
```diff
@@ -0,0 +1,802 @@
...
+                    "CANCELLED: $STATUS_DESCRIPTION_OF_INTERRUPTED_REQUEST",
+                )
+            )
+        }
+        @Test
+        override fun `server terminated intermediate session (retry true)`() {
+            val streamObserver = mock<StreamObserver<Response>> { }
+            val handlerBaton = Baton("handler")
+
+//            val clientServerBaton = Baton("client-server")
```
OptimumCode, line 514: Please, cleanup the commented parts of the code

README.md
Nikita-Smirnov-Exactpro: User can sepecify deadlock timeout, also, act component has 10 s default timeout.

OptimumCode: > act component has 10 s default timeout.

Who said that?) If it is a default value in our template it does not mean that all other act components have the same value.

And what is the `deadlock timeout`? I don't see such a configuration parameter

src/test/kotlin/com/exactpro/th2/common/schema/grpc/router/impl/DefaultGrpcRouterTest.kt
Nikita-Smirnov-Exactpro: I use fixed retry timeout in these tests. I wait time to grantee that client switch to retry mode. The time can be reduced because gRPC router begin retry after first failure

README.md
Nikita-Smirnov-Exactpro: Reduced to 1 minute

OptimumCode (approved):

README.md
Nikita-Smirnov-Exactpro: sorry deadline timeout )

PR #283, opened 2023-10-23
[TH2-2212] merged with dev 5

lumber1000:


src/main/java/com/exactpro/th2/common/schema/grpc/router/AbstractGrpcRouter.java
```diff
@@ -41,6 +41,7 @@
 import java.util.Objects;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ExecutorService;
```
OptimumCode, line 44: Duplicated import

README.md
```diff
@@ -604,6 +604,13 @@ dependencies {
 + Work was done to eliminate vulnerabilities in _common_ and _bom_ dependencies.
     + **th2-bom must be updated to _4.0.3_ or higher.**
 
+### 3.42.1
++ Added retry in case of a RabbitMQ channel or connection error (when possible).
++ Added InterruptedException to basicConsume method signature.
++ Added additional logging for RabbitMQ errors.
++ Fixed connection recovery delay time.
++ Integration tests for these scenarios.
```
OptimumCode, line 612: We merge into dev-version-5. These changes were not released for common v3. So, probably, we need to move correct the change log here

src/main/java/com/exactpro/th2/common/schema/message/impl/rabbitmq/connection/ConnectionManager.java
```diff
@@ -70,9 +77,8 @@ public class ConnectionManager implements AutoCloseable {
     public static final String EMPTY_ROUTING_KEY = "";
     private static final Logger LOGGER = LoggerFactory.getLogger(ConnectionManager.class);
 
-    private final Connection connection;
+    public final Connection connection;
```
OptimumCode, line 80: Why have we made the connection public? This is implementation details and nobody should have access to the connection field

src/main/java/com/exactpro/th2/common/schema/message/impl/rabbitmq/connection/ConnectionManager.java
```diff
@@ -215,54 +219,120 @@ private void turnOffReadiness(Throwable exception){
...
+    private void recoverSubscriptionsOfChannel(Channel channel) {
+        channelChecker.execute(() -> {
+            try {
+                var pinToChannelHolderOptional =
+                        channelsByPin
+                                .entrySet()
+                                .stream()
+                                .filter(entry -> Objects.nonNull(entry.getValue().channel) && channel.getChannelNumber() == entry.getValue().channel.getChannelNumber())
+                                .findAny();
+                if (pinToChannelHolderOptional.isPresent()) {
```
OptimumCode, line 267: Can we use `ifPresent` method here instead? Using an optional like this is no difference from getting the value with `orElse(null)` and checking that value for `null`. If we can't probably it is better to use `orElse(null)` method

src/main/java/com/exactpro/th2/common/schema/message/impl/rabbitmq/connection/ConnectionManager.java
```diff
@@ -281,18 +351,22 @@ public void close() {
             }
         }
 
+        for (ChannelHolder channelHolder: channelsByPin.values()) {
+            channelHolder.channel.abort();
+        }
```
OptimumCode, line 356: All channels should be closed automatically when we close the connection. Do we need to explicitly call `abort` method on the channel

src/main/java/com/exactpro/th2/common/schema/message/impl/rabbitmq/connection/ConnectionManager.java
```diff
@@ -460,7 +563,7 @@ private void waitForRecovery(ShutdownNotifier notifier) {
     }
 
     private boolean isConnectionRecovery(ShutdownNotifier notifier) {
-        return !notifier.isOpen() && !connectionIsClosed.get();
+        return !(notifier instanceof AutorecoveringChannel) && !notifier.isOpen() && !connectionIsClosed.get();
```
OptimumCode, line 566: Why have we added this check here?

src/main/kotlin/com/exactpro/th2/common/schema/message/impl/rabbitmq/configuration/RabbitMQConfiguration.kt
```diff
@@ -45,4 +47,49 @@ data class ConnectionManagerConfiguration(
...
+                retryTimeDeviationPercent
+            ))
+        }
+    }
+
+}
+
+data class RetryingDelay(val tryNumber: Int, val delay: Int) {
+    companion object {
+        fun getRecoveryDelay(
```
OptimumCode, line 67: Please, add `@JvmStatic` annotation for method in order to correctly use in from Java

src/test/kotlin/com/exactpro/th2/common/schema/message/impl/rabbitmq/AbstractRabbitRouterIntegrationTest.kt
```diff
@@ -203,7 +202,7 @@ class AbstractRabbitRouterIntegrationTest {
     companion object {
         private val K_LOGGER = KotlinLogging.logger { }
 
-        private const val RABBITMQ_MANAGEMENT_ALPINE = "rabbitmq:3.11.2-management-alpine"
+        private const val RABBITMQ_MANAGEMENT_ALPINE = "rabbitmq:3.8-management-alpine"
```
OptimumCode, line 205: Why did you downgrade the version?

src/main/java/com/exactpro/th2/common/schema/message/impl/rabbitmq/connection/ConnectionManager.java
lumber1000: I assume that we should wait for reconnection only for autorecovering connections. But looks like the condition is wrong. Probably correct condition is `(notifier instanceof AutorecoveringConnection) && !notifier.isOpen() && !connectionIsClosed.get()`.

src/main/java/com/exactpro/th2/common/schema/message/impl/rabbitmq/connection/ConnectionManager.java
```diff
@@ -605,6 +727,63 @@ public void withLock(boolean waitForRecovery, ChannelConsumer consumer) throws I
             }
         }
 
+        public void retryingPublishWithLock(ChannelConsumer consumer, ConnectionManagerConfiguration configuration) throws InterruptedException {
+            Iterator<RetryingDelay> iterator = configuration.createRetryingDelaySequence().iterator();
+            lock.lock();
+            try {
+                var currentValue = iterator.next();
```
OptimumCode, line 734: Shouldn't we do that inside the loop? From my understanding each retry should take the next value from that iterator

src/main/java/com/exactpro/th2/common/schema/message/impl/rabbitmq/connection/ConnectionManager.java
```diff
@@ -165,45 +185,29 @@ public void handleTopologyRecoveryException(Connection conn, Channel ch, Topolog
...
-                                    / (connectionManagerConfiguration.getMaxRecoveryAttempts() - 1)
-                                    * tmpCountTriesToRecovery
-                                : 0);
+                    int minTime = connectionManagerConfiguration.getMinConnectionRecoveryTimeout();
+                    int maxTime = connectionManagerConfiguration.getMaxConnectionRecoveryTimeout();
+                    int maxRecoveryAttempts = connectionManagerConfiguration.getMaxRecoveryAttempts();
+                    int deviationPercent = connectionManagerConfiguration.getRetryTimeDeviationPercent();
+
+                    LOGGER.debug("Try to recovery connection to RabbitMQ. Count tries = {}", recoveryAttempts);
+                    int recoveryDelay = RetryingDelay.Companion.getRecoveryDelay(recoveryAttempts, minTime, maxTime, maxRecoveryAttempts, deviationPercent);
```
OptimumCode, line 204: If you added `@JvmStatic` to the method you now can remove access to the `Companion` object from the method call

src/main/kotlin/com/exactpro/th2/common/schema/message/impl/rabbitmq/configuration/RabbitMQConfiguration.kt
```diff
@@ -37,12 +38,60 @@ data class ConnectionManagerConfiguration(
...
+            numberOfTries: Int,
+            minTime: Int,
+            maxTime: Int,
+            maxRecoveryAttempts: Int,
+            deviationPercent: Int
+        ): Int {
+            return if (numberOfTries <= maxRecoveryAttempts) {
+                getRecoveryDelayWithIncrement(numberOfTries, minTime, maxTime, maxRecoveryAttempts)
+            } else getRecoveryDelayWithDeviation(maxTime, deviationPercent)
```
OptimumCode, line 78: Please, add brackets to the `else` block as well. It makes the whole statement more readable and consistent

src/test/kotlin/com/exactpro/th2/common/schema/message/impl/rabbitmq/connection/TestConnectionManager.kt
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2022 Exactpro (Exactpro Systems Limited)
+ * Copyright 2023 Exactpro (Exactpro Systems Limited)
```
OptimumCode, line 2: I think we should update it to `2022-2023` because it is not a new class you have just added

src/test/kotlin/com/exactpro/th2/common/schema/message/impl/rabbitmq/connection/TestConnectionManager.kt
```diff
@@ -107,14 +109,726 @@ class TestConnectionManager {
...
+                username = container.adminUsername,
+                password = container.adminPassword,
+            ),
+            configuration
+        )
+
     @Test
+    @Disabled
+    // TODO: this test is no more relevant
+    // TODO: we need to change test scenario or remove it
```
OptimumCode, line 829: Could you please elaborate this? What exactly is not longer relevant?

src/test/kotlin/com/exactpro/th2/common/schema/message/ContainerConstants.kt
```diff
@@ -0,0 +1,16 @@
+package com.exactpro.th2.common.schema.message
```
OptimumCode, line 1: Please, add missing copyright

OptimumCode (changes requested):

src/main/java/com/exactpro/th2/common/schema/message/impl/rabbitmq/connection/ConnectionManager.java
```diff
@@ -215,54 +219,121 @@ private void turnOffReadiness(Throwable exception){
         try {
             connection = factory.newConnection();
             LOGGER.info("Created RabbitMQ connection {} [{}]", connection, connection.hashCode());
+            addShutdownListenerToConnection(this.connection);
+            addBlockedListenersToConnection(this.connection);
+            addRecoveryListenerToConnection(this.connection);
```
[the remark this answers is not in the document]
lumber1000: Yes it will, I've removed auto-recovery from config.

src/main/java/com/exactpro/th2/common/schema/message/impl/rabbitmq/connection/ConnectionManager.java
```diff
@@ -215,54 +219,121 @@ private void turnOffReadiness(Throwable exception){
...
+
+                var pinIdToChannelHolder = pinToChannelHolderOptional.orElse(null);
+                if (pinIdToChannelHolder != null) {
+                    PinId pinId = pinIdToChannelHolder.getKey();
+                    ChannelHolder channelHolder = pinIdToChannelHolder.getValue();
+
+                    SubscriptionCallbacks subscriptionCallbacks = channelHolder.subscriptionCallbacks;
+                    if (subscriptionCallbacks != null) {
+                        LOGGER.info("Changing channel for holder with pin id: " + pinId.toString());
+                        channelsByPin.remove(pinId);
```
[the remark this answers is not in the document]
lumber1000: we execute this code on single thread executor service

src/test/kotlin/com/exactpro/th2/common/schema/message/impl/rabbitmq/connection/TestConnectionManager.kt
lumber1000: This test expected exception, but now channel tries to recover. I've updated test.

src/test/resources/rabbitmq_it.conf
```diff
@@ -0,0 +1,2 @@
+consumer_timeout = 50000
```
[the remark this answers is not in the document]
lumber1000: I've reduced. But RabbitMQ checks this timeout for expiration only once per minute.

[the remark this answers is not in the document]
OptimumCode: Thank you for your explanation. Probably, in this case, we should set the 1-minute interval and leave a comment that explains why there is no point in using less timeout.

This is the part of the [official documentation](https://www.rabbitmq.com/consumers.html#acknowledgement-timeout):
_Whether the timeout should be enforced is evaluated periodically, at one minute intervals. **Values lower than one minute are not supported,** and values lower than five minutes are not recommended._

src/test/kotlin/com/exactpro/th2/common/schema/message/impl/rabbitmq/connection/TestConnectionManager.kt
```diff
@@ -824,29 +816,25 @@ class TestConnectionManager {
...
+            rabbitMQContainer.start()
+            LOGGER.info { "Started with port ${rabbitMQContainer.amqpPort}" }
+
+            createConnectionManager(rabbitMQContainer).use { firstManager ->
+                createConnectionManager(rabbitMQContainer).use { secondManager ->
+                    val queue = firstManager.queueDeclare()
+                    val consumerThread = thread { secondManager.basicConsume(queue, { _, _, _ -> }, {}) }
+                    consumerThread.join(5_000)
+                    val isAlive = consumerThread.isAlive
+                    consumerThread.stop()
```
OptimumCode, line 830: The `stop` method is deprecated in Java. We probably should correct the approach for testing this...

src/test/kotlin/com/exactpro/th2/common/schema/message/impl/rabbitmq/connection/TestConnectionManager.kt
```diff
@@ -824,29 +816,25 @@ class TestConnectionManager {
...
-                        extracted(firstManager, secondManager, queue, 6)
-                    }
+        RabbitMQContainer(RABBITMQ_IMAGE_NAME).use { rabbitMQContainer ->
+            rabbitMQContainer.start()
+            LOGGER.info { "Started with port ${rabbitMQContainer.amqpPort}" }
+
+            createConnectionManager(rabbitMQContainer).use { firstManager ->
+                createConnectionManager(rabbitMQContainer).use { secondManager ->
+                    val queue = firstManager.queueDeclare()
+                    val consumerThread = thread { secondManager.basicConsume(queue, { _, _, _ -> }, {}) }
```
OptimumCode, line 827: The attempt to subscribe to the queue from another connection causes:
_RabbitMQ soft error occupied: (reply-code=405, reply-text=RESOURCE_LOCKED - cannot obtain exclusive access to locked queue 'amq.gen-hdta456EkjDFt7Q_nqsprg' in vhost '/'. It could be originally declared on another connection or the exclusive property value does not match that of the original declaration., class-id=60, method-id=20) on channel AMQChannel(amqp://guest@127.0.0.1:32775/,1)_ error.

From my point of view, this looks like an unrecoverable error. According to [official documentation](https://www.rabbitmq.com/queues.html#exclusive-queues) the exclusive queue can only be managed by the connection it was created by. We must propagate the error back to the consumer.

@Nikita-Smirnov-Exactpro Do you have any objection to this?

OptimumCode (changes requested):

src/main/java/com/exactpro/th2/common/schema/message/impl/rabbitmq/connection/ConnectionManager.java
```diff
@@ -215,54 +219,121 @@ private void turnOffReadiness(Throwable exception){
...
+                        basicConsume(pinId.queue, subscriptionCallbacks.deliverCallback, subscriptionCallbacks.cancelCallback);
+                    }
+                }
+            } catch (IOException | InterruptedException e) {
+                LOGGER.warn("Exception during channel's subscriptions recovery", e);
+                throw new RuntimeException(e);
```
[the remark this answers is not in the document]
OptimumCode: Do you have any updates regarding this change request? We should at least set the interrupt flag in case of InterruptedException

src/main/java/com/exactpro/th2/common/schema/message/impl/rabbitmq/connection/ConnectionManager.java
```diff
@@ -441,6 +449,7 @@ public void confirm() throws IOException {
                     }
                 }, cancelCallback), configuration);
 
+        holder.isSubscribed = true;
```
OptimumCode, line 452: Can we encapsulate this into the holder class? I think it would be better to have just a `subscribed` method in holder class

src/test/kotlin/com/exactpro/th2/common/schema/message/impl/rabbitmq/connection/TestConnectionManager.kt
Nikita-Smirnov-Exactpro: Common generates unique name for exclusive queue. The error above is critical for the current implementation.
Also, some messages can be missed if we recreate exclusive queue with the same name.
The problem should be critical as result

Nikita-Smirnov-Exactpro: Critical error in this case looks like a concept problem and we should think about solving in future

src/main/java/com/exactpro/th2/common/schema/message/impl/rabbitmq/connection/ConnectionManager.java
```diff
@@ -276,15 +276,14 @@ private void recoverSubscriptionsOfChannel(int channelNumber) {
...
+                        if (channelHolder.subscribed()) {
                             LOGGER.info("Changing channel for holder with pin id: " + pinId.toString());
                             basicConsume(pinId.queue, subscriptionCallbacks.deliverCallback, subscriptionCallbacks.cancelCallback);
                         }
                     }
                 }
             } catch (IOException e) {
-                LOGGER.warn("Exception during channel's subscriptions recovery", e);
-                throw new RuntimeException(e);
+                LOGGER.warn("Failed to recovery channel's subscriptions", e);
```
OptimumCode, line 286: Why don't we throw an exception anymore? Could you please add a comment about why it is okay here?
Will we try to resubscribe again if it is a temporal error?

src/main/java/com/exactpro/th2/common/schema/message/impl/rabbitmq/connection/ConnectionManager.java
```diff
@@ -605,6 +728,85 @@ public void withLock(boolean waitForRecovery, ChannelConsumer consumer) throws I
...
+                        break;
+                    } catch (IOException | ShutdownSignalException e) {
+                        var currentValue = iterator.next();
+                        int recoveryDelay = currentValue.getDelay();
+                        LOGGER.warn("Retrying publishing #{}, waiting for {}ms. Reason: {}", currentValue.getTryNumber(), recoveryDelay, e);
+                        TimeUnit.MILLISECONDS.sleep(recoveryDelay);
+
+                        // we should not recover channel if it's connection is closed, because if we'll do it channel
+                        // will be also auto recovered by RabbitMQ client during recovering of connection, and we'll
+                        // get two new channels instead of one closed.
```
OptimumCode, line 748:
```suggestion
                        // We should not recover the channel if its connection is closed
                        // If we do that the channel will be also auto recovered by RabbitMQ client
                        // during connection recovery, and we will get two new channels instead of one closed.
```

OptimumCode (approved):

src/main/java/com/exactpro/th2/common/schema/message/impl/rabbitmq/connection/ConnectionManager.java
```diff
@@ -247,8 +247,14 @@ private void addShutdownListenerToChannel(Channel channel, Boolean withRecovery)
                         errorBuilder.append(channelCause);
                         String errorString = errorBuilder.toString();
                         LOGGER.warn(errorString);
-                        if (withRecovery && errorString.contains("PRECONDITION_FAILED")) {
-                            recoverSubscriptionsOfChannel(channel.getChannelNumber());
+                        if (withRecovery &&
+                                (errorString.contains("reply-text=PRECONDITION_FAILED")
```
OptimumCode, line 251: I think we should move these error texts into a static list and create a method to check if the error is recoverable by iterating over this list and checking if error text contains expected substrings

src/main/java/com/exactpro/th2/common/schema/message/impl/rabbitmq/connection/ConnectionManager.java
```diff
@@ -681,6 +687,8 @@ private static class ChannelHolder {
         private Channel channel;
         @GuardedBy("lock")
         private boolean isSubscribed = false;
+        @GuardedBy("lock")
+        private boolean isSubscribing = false;
```
OptimumCode, line 691: Maybe we should replace this and `isSubscribed` boolean variables with enum that will have channel states

src/test/kotlin/com/exactpro/th2/common/schema/message/impl/rabbitmq/connection/TestConnectionManager.kt
```diff
@@ -445,8 +444,12 @@ class TestConnectionManager {
                     LOGGER.info { "Rabbit address- ${it.host}:${it.amqpPort}" }
 
                     LOGGER.info { "Restarting the container" }
-                    restartContainer(it)
-                    Thread.sleep(5000)
+                    it.stop()
+                    Thread.sleep(5_000)
```
OptimumCode, line 448: Why do we need a delay between stop and start? As far as I remember the method won't return until the container is finally stopped. The same about `start` method

src/main/java/com/exactpro/th2/common/schema/message/impl/rabbitmq/connection/ConnectionManager.java
```diff
@@ -247,14 +252,12 @@ private void addShutdownListenerToChannel(Channel channel, Boolean withRecovery)
...
                         LOGGER.warn(errorString);
-                        if (withRecovery &&
-                                (errorString.contains("reply-text=PRECONDITION_FAILED")
-                                || errorString.contains("reply-text=NOT_FOUND"))
-                        ) {
+                        if (withRecovery && recoverableErrors.stream().anyMatch(errorString::contains)) {
                             int channelNumber =  channel.getChannelNumber();
                             var pinIdToChannelHolder = getChannelHolderByChannelNumber(channelNumber);
-                            if (pinIdToChannelHolder != null && !pinIdToChannelHolder.getValue().subscribing())
+                            if ((pinIdToChannelHolder != null) && (pinIdToChannelHolder.getValue().state() != ChannelState.SUBSCRIBING)) {
```
OptimumCode, line 258: I don't think that it is okay that we have access to the internal state without a lock. When I suggested using enum instead of two booleans I meant that we should do that inside the holder. All interactions outside should be left as they were. Also, the work with the state should be done under the lock

OptimumCode (changes requested):

src/main/java/com/exactpro/th2/common/schema/message/impl/rabbitmq/connection/ConnectionManager.java
```diff
@@ -252,44 +252,56 @@ private void addShutdownListenerToChannel(Channel channel, Boolean withRecovery)
...
         });
     }
 
-    private @Nullable Map.Entry<PinId, ChannelHolder> getChannelHolderByChannelNumber(int channelNumber) {
+    private @Nullable Map.Entry<PinId, ChannelHolder> getChannelHolderByChannel(Channel channel) {
         return channelsByPin
                 .entrySet()
                 .stream()
-                .filter(entry -> Objects.nonNull(entry.getValue().channel) && channelNumber == entry.getValue().channel.getChannelNumber())
+                .filter(entry -> channel == entry.getValue().channel)
```
OptimumCode, line 274: @lumber1000 could you please advise why did we go back to using the channel here? If this is related to some edge cases or rabbit behavior please leave a comment in the code

src/main/java/com/exactpro/th2/common/schema/message/impl/rabbitmq/connection/ConnectionManager.java
```diff
@@ -252,44 +252,56 @@ private void addShutdownListenerToChannel(Channel channel, Boolean withRecovery)
...
 
-    private void recoverSubscriptionsOfChannel(int channelNumber) {
+    private void recoverSubscriptionsOfChannel(@NotNull final PinId pinId, Channel channel, @NotNull final ChannelHolder holder) {
         channelChecker.execute(() -> {
             try {
-                var pinIdToChannelHolder = getChannelHolderByChannelNumber(channelNumber);
-                if (pinIdToChannelHolder != null) {
-                    PinId pinId = pinIdToChannelHolder.getKey();
-                    ChannelHolder channelHolder = pinIdToChannelHolder.getValue();
+                holder.lock.lock();
```
OptimumCode, line 282: The lock should be taken outside the try-finally block. Otherwise, we might call unlock if the locking was not successful

src/main/java/com/exactpro/th2/common/schema/message/impl/rabbitmq/connection/ConnectionManager.java
```diff
@@ -252,44 +252,56 @@ private void addShutdownListenerToChannel(Channel channel, Boolean withRecovery)
...
+                        return;
+                    }
 
-                    SubscriptionCallbacks subscriptionCallbacks = channelHolder.subscriptionCallbacks;
                     if (subscriptionCallbacks != null) {
                         channelsByPin.remove(pinId);
-                        if (channelHolder.state == ChannelState.SUBSCRIBED) {
-                            LOGGER.info("Changing channel for holder with pin id: " + pinId.toString());
-                            basicConsume(pinId.queue, subscriptionCallbacks.deliverCallback, subscriptionCallbacks.cancelCallback);
+                        if (holder.isSubscribed) {
```
OptimumCode, line 293: Why do we operate on the internal holder state? In my understanding the lock and responsibility to track whether the channel is subscribing or not is fully on the channel's holder. But right now we manage that outside the holder. Could you please advise on why have you decided to go this way?

src/main/java/com/exactpro/th2/common/schema/message/impl/rabbitmq/connection/ConnectionManager.java
lumber1000: Because we need holder for exact channel object, not for "channel number N", if channel was changed in holder we don't need to recover it.

OptimumCode: Got it, thank you. Feel free to leave that as a comment in the code if you think it might help to prevent problems in the future if anybody is going to change the code in this method

src/test/kotlin/com/exactpro/th2/common/schema/grpc/router/impl/DefaultGrpcRouterTest.kt
```diff
@@ -289,10 +289,6 @@ internal class DefaultGrpcRouterTest {
             )
         }
 
-        /* FIXME: this test sometimes failing with TimeoutException
```
OptimumCode: Is the test fixed after merging with dev-release-5? Just checking if it was not removed accidentally

src/test/kotlin/com/exactpro/th2/common/schema/message/impl/rabbitmq/connection/TestConnectionManager.kt
```diff
@@ -707,6 +709,190 @@ class TestConnectionManager {
             }
     }
 
+    @Test
```
OptimumCode, line 712: I think you might consider using TestFactory for this. It allows to creation of multiple tests from a single method. Not sure if it fits into the requirements you need but you might take a look and decide for yourself

src/test/kotlin/com/exactpro/th2/common/schema/message/impl/rabbitmq/connection/TestConnectionManager.kt
```diff
@@ -707,6 +709,190 @@ class TestConnectionManager {
...
+                    repeat(params.messagesToSend) {
+                        LOGGER.info { "Sending message ${it + 1} to queue ${params.queueName}" }
+                        putMessageInQueue(container, params.queueName)
+                    }
+
+                    TestCaseContext(connectionManager, consumersThreads, consumerCounters)
+                }
+
+                LOGGER.info { "Sleeping..." }
+                Thread.sleep(63000)
```
OptimumCode, line 862: Could you please advise why 63 seconds exactly? Maybe we need to rethink the test structure here to avoid such long timeouts. Also, is there any guarantee that 63 seconds is enough?

OptimumCode (approved):

src/main/java/com/exactpro/th2/common/schema/message/impl/rabbitmq/connection/ConnectionManager.java
OptimumCode: You are right, sorry

OptimumCode:
@Nikita-Smirnov-Exactpro Could you please check the PR and if you don't have any objections I think we can merge it

src/test/kotlin/com/exactpro/th2/common/schema/grpc/router/impl/DefaultGrpcRouterTest.kt
lumber1000: I can't reproduce this problem now.

src/test/kotlin/com/exactpro/th2/common/schema/message/impl/rabbitmq/connection/TestConnectionManager.kt
lumber1000: I wrote this method to allow to  execute multiple tests in parallel because of long delay needed in this test.

src/test/kotlin/com/exactpro/th2/common/schema/message/impl/rabbitmq/connection/TestConnectionManager.kt
lumber1000: We need to wait 1 second for consumer timeout expiration. Then we need to wait until rabbitmq checks channels for timeout, it performs such check once a minute, so we need to wait for 60+ seconds. And then we need to wait for a while for recovering of connection and resubscribing channels. If 63 seconds will not be enough (e.g. in newer versions of rabbitmq) the test will fail, and we will fix it.

src/test/kotlin/com/exactpro/th2/common/schema/grpc/router/impl/DefaultGrpcRouterTest.kt
OptimumCode: Understood, thank you

src/test/kotlin/com/exactpro/th2/common/schema/message/impl/rabbitmq/connection/TestConnectionManager.kt
OptimumCode: Thank you for your explanation. I think that would be helpful for "future you" or me or anybody else if we leave a comment about that in the test. Because I don't believe it is possible to always remember this specific rabbitmq behavior and our internal logic

src/main/kotlin/com/exactpro/th2/common/schema/message/impl/rabbitmq/configuration/RabbitMQConfiguration.kt
```diff
@@ -18,31 +18,84 @@ package com.exactpro.th2.common.schema.message.impl.rabbitmq.configuration
...
+            minTime: Int,
+            maxTime: Int,
+            maxRecoveryAttempts: Int,
+            deviationPercent: Int
+        ): Int {
+            return if (numberOfTries <= maxRecoveryAttempts) {
+                getRecoveryDelayWithIncrement(numberOfTries, minTime, maxTime, maxRecoveryAttempts)
+            } else {
+                getRecoveryDelayWithDeviation(maxTime, deviationPercent)
+            }
```
Nikita-Smirnov-Exactpro, line 82: Maybe use divination in all cases because when several components lost connection at the same time, after recovery RabbitMQ will receive several reconnect requests at the same time.
It is not critical problem, but we should keep in mind

src/main/java/com/exactpro/th2/common/schema/message/impl/rabbitmq/connection/ConnectionManager.java
```diff
@@ -215,61 +221,140 @@ private void turnOffReadiness(Throwable exception){
...
-                public void handleRecovery(Recoverable recoverable) {
-                    LOGGER.debug("Count tries to recovery connection reset to 0");
-                    connectionRecoveryAttempts.set(0);
-                    metrics.getReadinessMonitor().enable();
-                    LOGGER.debug("Set RabbitMQ readiness to true");
+    private void recoverSubscriptionsOfChannel(@NotNull final PinId pinId, Channel channel, @NotNull final ChannelHolder holder) {
+        channelChecker.execute(() ->
+            holder.getCallbacksForRecovery(channel).ifPresent(subscriptionCallbacks -> {
+                LOGGER.info("Changing channel for holder with pin id: " + pinId);
+                channelsByPin.remove(pinId);
```
Nikita-Smirnov-Exactpro, line 284: Maybe we should close broken channel or do the similar action to be sure that resources is closed

src/main/java/com/exactpro/th2/common/schema/message/impl/rabbitmq/connection/ConnectionManager.java
```diff
@@ -215,61 +221,140 @@ private void turnOffReadiness(Throwable exception){
...
+    public void close()  throws IOException {
         if (connectionIsClosed.getAndSet(true)) {
             LOGGER.info("Connection manager already closed");
             return;
         }
 
         LOGGER.info("Closing connection manager");
 
+        for (ChannelHolder channelHolder: channelsByPin.values()) {
+            channelHolder.channel.abort();
```
Nikita-Smirnov-Exactpro, line 355: We doesn't execute the code below if this instruction throws the IOException.
Maybe collect all exceptions and throw an aggregated exception with the thrown exceptions as suppressed

src/main/java/com/exactpro/th2/common/schema/message/impl/rabbitmq/connection/ConnectionManager.java
```diff
@@ -605,6 +731,107 @@ public void withLock(boolean waitForRecovery, ChannelConsumer consumer) throws I
...
+                        tempChannel = recreateChannel();
+                    }
+
+                    var tag = mapper.map(tempChannel);
+                    isSubscribed = true;
+                    return tag;
+                } catch (IOException e) {
+                    var reason = tempChannel.getCloseReason();
+                    isChannelClosed = reason != null;
+                    if (isChannelClosed && reason.getMessage().contains("reply-text=RESOURCE_LOCKED")) {
```
Nikita-Smirnov-Exactpro, line 782: Why we shouldn't retry in this case?
Our API inconsistent
retryingPublishWithLock(...) throws InterruptedException
retryingConsumeWithLock(...) throws InterruptedException, IOException

src/main/java/com/exactpro/th2/common/schema/message/impl/rabbitmq/connection/ConnectionManager.java
```diff
@@ -566,12 +675,15 @@ private static class ChannelHolder {
         private final Supplier<Channel> supplier;
         private final BiConsumer<ShutdownNotifier, Boolean> reconnectionChecker;
         private final int maxCount;
+        private final SubscriptionCallbacks subscriptionCallbacks;
```
Nikita-Smirnov-Exactpro, line 678: Please rename to SubscriptionCallback because it contains only one call back now

src/main/java/com/exactpro/th2/common/schema/message/impl/rabbitmq/connection/ConnectionManager.java
```diff
@@ -605,6 +731,107 @@ public void withLock(boolean waitForRecovery, ChannelConsumer consumer) throws I
...
+                iterator = handleAndSleep(configuration, iterator, "Retrying consume", exception);
+            }
+        }
+
+        public Optional<SubscriptionCallbacks> getCallbacksForRecovery(Channel channelToRecover) {
+            lock.lock();
+            SubscriptionCallbacks callbacks = null;
+            try {
+                if (channel != channelToRecover) {
+                    LOGGER.warn("Channel already recovered");
```
Nikita-Smirnov-Exactpro, line 798: This case looks like we have bug in our code.
Maybe throw an exception in this case

src/main/java/com/exactpro/th2/common/schema/message/impl/rabbitmq/connection/ConnectionManager.java
OptimumCode: @Nikita-Smirnov-Exactpro We should not retry in this case because we never will be able to connect to the queue if we receive this error. This error happens if we try to subscribe to an exclusive queue that was created by another connection.
In the previous discussions in this PR we agreed that we should report this and do no try to recover

Nikita-Smirnov-Exactpro: Please add comment for this case

src/main/java/com/exactpro/th2/common/schema/message/impl/rabbitmq/connection/ConnectionManager.java
lumber1000: but it actually contains two callbacks: deliver callback and cancel callback

Nikita-Smirnov-Exactpro: OK, I missed this fact

src/main/java/com/exactpro/th2/common/schema/message/impl/rabbitmq/connection/ConnectionManager.java
lumber1000: I believe the channel is already closed when this code is executed, because this code is called from shutdown listener.

src/main/java/com/exactpro/th2/common/schema/message/impl/rabbitmq/connection/ConnectionManager.java
lumber1000: I think we should just log exceptions the same as for channel.close(). Currently ConnectionManager::close does not throws IOException, so I just catch and log it.

src/main/java/com/exactpro/th2/common/schema/message/impl/rabbitmq/connection/ConnectionManager.java
lumber1000: Exception will not be handled because this code executed in executor service.

src/main/java/com/exactpro/th2/common/schema/message/impl/rabbitmq/AbstractRabbitSubscriber.java
```diff
@@ -189,6 +189,10 @@ private SubscriberMonitor basicConsume() {
             return connectionManager.basicConsume(queue, this::handle, this::canceled);
         } catch (IOException e) {
             throw new IllegalStateException("Can not subscribe to queue = " + queue, e);
+        } catch (InterruptedException e) {
```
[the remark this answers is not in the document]
lumber1000: If we are in catch block subscribe retry process is already interrupted (because it's in try block - connectionManager.basicConsume).

src/test/kotlin/com/exactpro/th2/common/schema/message/impl/rabbitmq/connection/TestConnectionManager.kt
```diff
@@ -893,79 +873,6 @@ class TestConnectionManager {
             }
     }
 
-    @Test
-    fun `connection manager handles ack timeout and subscription cancel`() {
```
OptimumCode: @lumber1000 could you please advise why have you removed this test?

OptimumCode (approved):

src/main/java/com/exactpro/th2/common/schema/message/impl/rabbitmq/connection/ConnectionManager.java
Nikita-Smirnov-Exactpro: Entry point of our tasks should have try ... catch block to catch all Exceptions.

src/main/java/com/exactpro/th2/common/schema/message/impl/rabbitmq/connection/ConnectionManager.java
Nikita-Smirnov-Exactpro: let's add the close logic just in case

src/test/kotlin/com/exactpro/th2/common/schema/message/impl/rabbitmq/connection/TestConnectionManager.kt
lumber1000: this test is included into `connection manager handles ack timeout (multiple subscribers in parallel)` as second test case ("queue2").

src/main/java/com/exactpro/th2/common/schema/message/impl/rabbitmq/connection/ConnectionManager.java
```diff
@@ -732,30 +735,30 @@ public void withLock(boolean waitForRecovery, ChannelConsumer consumer) throws I
...
+                    if (!tempChannel.isOpen() && tempChannel.getConnection().isOpen()) {
+                        tempChannel = recreateChannel();
                     }
+                } finally {
+                    lock.unlock();
                 }
-            } finally {
-                lock.unlock();
+
+                TimeUnit.MILLISECONDS.sleep(recoveryDelay);
```
Nikita-Smirnov-Exactpro, line 761: Why this instruction was moved from catch block?

lumber1000: To avoid locking for a long time.

Nikita-Smirnov-Exactpro: all operation with a channel are executed under lock. If one operation faces to a retry problem another will wait too.
In this case we can block another operations until resolve the retry problem

lumber1000: ok, i will add one more lock for processing callbacks.

Nikita-Smirnov-Exactpro: No, please satay as is

Nikita-Smirnov-Exactpro: Several publishers follow replay schedule in your case.

src/main/java/com/exactpro/th2/common/schema/message/impl/rabbitmq/connection/ConnectionManager.java
```diff
@@ -279,20 +278,24 @@ private void addShutdownListenerToChannel(Channel channel, Boolean withRecovery)
...
-                channelsByPin.remove(pinId);
-                try {
+        channelChecker.execute(() -> {
+            try {
+                var subscriptionCallbacks = holder.getCallbacksForRecovery(channel);
+
+                if (subscriptionCallbacks != null) {
+                    LOGGER.info("Changing channel for holder with pin id: " + pinId);
+                    channelsByPin.remove(pinId);
+                    if (channel.isOpen()) channel.abort();
```
Nikita-Smirnov-Exactpro, line 288: if (removedHolder == holder) {
                       if (channel.isOpen()) channel.abort();
                    } else {
                       ???
                    }

src/main/java/com/exactpro/th2/common/schema/message/impl/rabbitmq/connection/ConnectionManager.java
lumber1000: we need 2nd lock to read `channel` and `isSubscribed` values to filter out unnecessary callbacks without waiting for channel recovery. Otherwise all callbacks will be accumulated in executor service queue until the connection and channels will be recovered (and every recovery attempt generates new callback).

Nikita-Smirnov-Exactpro: We can use `volatile` or AtomicBoolean instead. Less lock, simple code, less chance to make a deadlock in future changes

OptimumCode:
@Nikita-Smirnov-Exactpro @lumber1000 Hi, what are we doing about this one? If we have tested it let's pull the latest change and if integration tests pass merge the PR

Nikita-Smirnov-Exactpro:
> @Nikita-Smirnov-Exactpro @lumber1000 Hi, what are we doing about this one? If we have tested it let's pull the latest change and if integration tests pass merge the PR

I agree with you, because we haven't got other tests

OptimumCode (approved):

PR #284, opened 2023-10-23
[TH2-5104] Use thread-safe initialization for resource in common factory

OptimumCode:


src/main/kotlin/com/exactpro/th2/common/schema/factory/LazyProvider.kt
```diff
@@ -0,0 +1,119 @@
...
+            val holder = State.Hold(initialize())
+            if (!reference.compareAndSet(State.Init, holder)) {
+                onClose.consume(holder.value)
+                error("provider '$name' already closed")
+            }
+            holder
+        } else {
+            var currentState: State<T>?
+            do {
+                currentState = reference.get()
```
Nikita-Smirnov-Exactpro, line 58: Maybe add Thread.yield() call

src/main/kotlin/com/exactpro/th2/common/schema/factory/LazyProvider.kt
```diff
@@ -0,0 +1,119 @@
...
+        }
+    }
+
+    @Throws(Exception::class)
+    override fun close() {
+        if (reference.compareAndSet(null, State.Closed)) {
+            // factory was not yet initialized. Just return
+            return
+        }
+
```
Nikita-Smirnov-Exactpro, line 70: I think we should add State.Init handling

OptimumCode: This is handled in `get` method already

Nikita-Smirnov-Exactpro: Sorry, but I don't see `get` method call in this method

OptimumCode: Because we don't need to call it here. If somebody calls `get` and `close` method concurrently and `close` method is invoked when the current state is `Init` the `get` method will handle resource closing because the resource is not yet created to close it in `close` method

Nikita-Smirnov-Exactpro (approved):

PR #286, opened 2023-12-06
[TH2-5143] Event builder checks is the book name from attached messag…

Nikita-Smirnov-Exactpro:
…e match to the event

### DIFF
--- a/src/main/java/com/exactpro/th2/common/event/Event.java
+++ b/src/main/java/com/exactpro/th2/common/event/Event.java
@@ -365,29 +365,31 @@ public class Event {
                 .setEndTimestamp(toTimestamp(endTimestamp))
                 .setStatus(getAggregatedStatus().eventStatus)
                 .setBody(ByteString.copyFrom(buildBody()));
+        List<String> problems = new ArrayList<>();
         if (parentId != null) {
             if (!Objects.equals(parentId.getBookName(), eventId.getBookName())) {
-                throw new IllegalArgumentException(
-                        "Book name mismatch in '" + toJson(parentId) + "' parent and '" +
-                                toJson(eventId) + "' current event ids"
-                );
+                problems.add("Book name mismatch in '" + toJson(parentId) + "' parent event id");
             }
             if (!Objects.equals(parentId.getScope(), eventId.getScope())) {
-                throw new IllegalArgumentException(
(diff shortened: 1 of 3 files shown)


src/main/java/com/exactpro/th2/common/event/Event.java
```diff
@@ -364,10 +366,22 @@ private com.exactpro.th2.common.grpc.Event toProto(
                 .setStatus(getAggregatedStatus().eventStatus)
                 .setBody(ByteString.copyFrom(buildBody()));
         if (parentId != null) {
+            if (!Objects.equals(parentId.getBookName(), eventId.getBookName())) {
```
OptimumCode, line 369: I think we should check the scope as well. Otherwise, it won't be possible to display parent and child events at the same time

src/test/kotlin/com/exactpro/th2/common/event/TestEvent.kt
```diff
@@ -36,6 +39,7 @@ import java.time.LocalDate
 import java.time.LocalDateTime
 import java.time.LocalTime
 import java.time.ZoneOffset
+import kotlin.IllegalArgumentException as IllegalArgumentException1
```
OptimumCode, line 42: Why do you need that? Kotlin's IllegalArgumentException is the alias to Java's IllegalArgumentException. It looks like we are doing something wrong here

OptimumCode (approved):

src/main/java/com/exactpro/th2/common/event/Event.java
OptimumCode, line 1: If we could combine all problems in one exception it would be a lot simpler for users to understand what is wrong. But we can keep the current state I think

OptimumCode (approved):

src/test/kotlin/com/exactpro/th2/common/event/TestEvent.kt
```diff
@@ -305,6 +309,42 @@ class TestEvent {
         )
     }
 
+    @Test
+    fun `book mismatch between attached message and event`() {
```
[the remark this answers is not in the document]
Nikita-Smirnov-Exactpro: Test for event with different book and scope in event and parent can't be written because Event.setParentId is private.
User define book and scope via to... methods

PR #287, opened 2024-01-03
Fix NPE in case the resource is AutoClosable and can be null

OptimumCode:


Nikita-Smirnov-Exactpro:
Please add test for problem case

OptimumCode:
@Nikita-Smirnov-Exactpro Added a test

PR #288, opened 2024-01-26
Th2 5157 provide not wildcard filter for mq routing

Nikita-Smirnov-Exactpro:


src/main/kotlin/com/exactpro/th2/common/schema/message/configuration/MessageRouterConfiguration.kt
```diff
@@ -92,7 +92,8 @@ enum class FieldFilterOperation {
     NOT_EQUAL,
     EMPTY,
     NOT_EMPTY,
-    WILDCARD
+    WILDCARD,
+    NOT_WILDCARD,
```
OptimumCode, line 96: This may not be enough. I see that there is a [JSON schema](https://github.com/th2-net/th2-infra/blob/master/chart/crds/th2-box-crd.yaml#L406) in the infra repo which describes the operation field as an enum.
We should check if this schema is actually applied and modify it as well if required

Nikita-Smirnov-Exactpro: Thank you for remind, as I remember from the last discussion we going to provide ability for customizing CRD for users.
I will add change into th2-infra repository too

.github/workflows/build-dev-release.yml
```diff
@@ -1,17 +1,14 @@
-name: Build and publish dev-release Java distributions to sonatype.
+name: Build and publish dev release jar to sonatype repository
 
-on:
-  push:
-    tags:
-      - \d+.\d+.\d+-dev
+on: workflow_dispatch
```
OptimumCode, line 3: You won't be able to execute this workflow until it is merged into the default branch. So, we either should keep trigger on tag push or make the `dev-version-5` branch a default branch

OptimumCode (approved):

PR #294, opened 2024-04-12
Use box name from `box.json` config as RabbitMQ connection name

Nikita-Smirnov-Exactpro:


src/main/java/com/exactpro/th2/common/schema/message/impl/rabbitmq/connection/ConnectionManager.java
```diff
@@ -218,7 +218,7 @@ private void turnOffReadiness(Throwable exception) {
         factory.setSharedExecutor(sharedExecutor);
 
         try {
-            connection = factory.newConnection();
+            connection = factory.newConnection(connectionName);
```
OptimumCode, line 221: note: We can have duplicated connection names in case we scale the component. There will be multiple pods with the same box name. therefore the connection name will be the same. How does rabbitmq react in such cases?

OptimumCode: I have checked this locally - RabbitMQ does no report any errors in this case and simply displays two connections with same name in UI

OptimumCode (changes requested):
issue: Please, increment the version in gradle.properties file

OptimumCode (dismissed):

Nikita-Smirnov-Exactpro:
> issue: Please, increment the version in gradle.properties file

Bumped version

OptimumCode (approved):

PR #295, opened 2024-04-15
Rise `th2_component` metric with box name as `name` label value

Nikita-Smirnov-Exactpro:


OptimumCode:
Are you sure that you based this PR on the right branch? I see comments that are not related to this PR

Nikita-Smirnov-Exactpro:
> Are you sure that you based this PR on the right branch? I see comments that are not related to this PR

Corrected, thank you

OptimumCode:
> Corrected, thank you

What is corrected? There are still commits that do not belong to this change. I think the only way to fix this is to rebase this branch based on the current master state (and drop all the duplicated commits)

OptimumCode (approved):
If you want you can merge it as is. But I think that we need to fix commit history first

Nikita-Smirnov-Exactpro:
> If you want you can merge it as is. But I think that we need to fix commit history first

I'm going to use squash and merge

PR #307, opened 2024-05-28
Updated kubernetes-client: `6.12.1`

Nikita-Smirnov-Exactpro:


build.gradle
```diff
@@ -138,17 +138,17 @@ dependencies {
     implementation 'io.prometheus:simpleclient_httpserver'
     implementation 'io.prometheus:simpleclient_log4j2'
 
-    implementation("com.squareup.okio:okio:3.8.0") {
+    implementation("com.squareup.okio:okio:3.4.0") {
```
OptimumCode, line 141: thought: we should leave a note about why we downgraded the library versions and we should update them once we migrate to kotlin 1.9

src/test/kotlin/com/exactpro/th2/common/util/TestMultiMapDeserializer.kt
```diff
@@ -32,23 +34,38 @@ class TestMultiMapDeserializer {
...
 
         init {
-            OBJECT_MAPPER.registerModule(KotlinModule())
+            OBJECT_MAPPER.registerModule(
+                KotlinModule.Builder()
+                    .withReflectionCacheSize(512)
+                    .configure(KotlinFeature.NullToEmptyCollection, false)
+                    .configure(KotlinFeature.NullToEmptyMap, false)
+                    .configure(KotlinFeature.NullIsSameAsDefault, false)
+                    .configure(KotlinFeature.SingletonSupport, false)
```
OptimumCode, line 43: issue (non-blocking): This might result it several instances of Kotlin `object` that is not correct. We should set enable this feature. Even if this is just in test

OptimumCode (approved):

PR #315, opened 2024-06-12
Provided ability to set either of raw body of several dody data to `E…

Nikita-Smirnov-Exactpro:
…vent` builder

### DIFF
--- a/src/main/java/com/exactpro/th2/common/event/Event.java
+++ b/src/main/java/com/exactpro/th2/common/event/Event.java
@@ -27,12 +27,14 @@ import com.fasterxml.jackson.datatype.jsr310.JavaTimeModule;
 import com.fasterxml.jackson.module.kotlin.KotlinFeature;
 import com.fasterxml.jackson.module.kotlin.KotlinModule;
 import com.google.protobuf.ByteString;
+import com.google.protobuf.UnsafeByteOperations;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.io.IOException;
+import java.nio.charset.StandardCharsets;
 import java.time.Instant;
 import java.util.ArrayList;
 import java.util.Collection;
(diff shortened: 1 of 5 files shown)


src/test/kotlin/com/exactpro/th2/common/event/TestEvent.kt
```diff
@@ -52,6 +52,7 @@ class TestEvent {
     private val data = EventUtils.createMessageBean("0123456789".repeat(20))
     private val dataSize = MAPPER.writeValueAsBytes(listOf(data)).size
     private val bigData = EventUtils.createMessageBean("0123456789".repeat(30))
+    private val rawBody = "Test raw data longer than 25 characters"
```
OptimumCode, line 55: question: What is it about 25 characters? If it has any meaning this should be explained. Otherwise, I believe, we should not leave text that causes questions when you look at it

Nikita-Smirnov-Exactpro: Added description for rawBody variable

src/main/java/com/exactpro/th2/common/event/Event.java
```diff
@@ -559,13 +598,21 @@ public Instant getEndTimestamp() {
...
-        return OBJECT_MAPPER.get().writeValueAsBytes(body);
+        if (rawBody == null) {
+            return OBJECT_MAPPER.get().writeValueAsBytes(body);
+        } else {
+            return rawBody;
+        }
     }
 
     protected String formatStateException(String fieldName, Object value) {
         return fieldName + " in event '" + id + "' already sed with value '" + value + '\'';
```
OptimumCode, line 609:
```suggestion
        return fieldName + " in event '" + id + "' already set with value '" + value + '\'';
```

src/main/java/com/exactpro/th2/common/event/Event.java
```diff
@@ -559,13 +598,21 @@ public Instant getEndTimestamp() {
...
     }
 
+    protected String formatRawBodyStateException(String fieldName) {
+        return fieldName + " can't be added to body data of event '" + id + "' because raw body is already sed";
```
OptimumCode, line 613:
```suggestion
        return fieldName + " can't be added to body data of event '" + id + "' because raw body is already set";
```

OptimumCode (approved):

OptimumCode (commented):
Please, update tests

OptimumCode (approved):

PR #325, opened 2024-09-26
Provided windows compatibility for ExactproMetaInf

Nikita-Smirnov-Exactpro:


src/main/kotlin/com/exactpro/th2/common/schema/factory/ExactproMetaInf.kt
```diff
@@ -106,10 +106,21 @@ internal class ExactproMetaInf(
...
                 return null
             }
         }
+
+        private fun URL.toPath(): Path = toURI().run {
+            // the code below are added to provide windows compatibility
+            when (scheme) {
+                "file" -> Path.of(this)
+                "jar" -> Path.of(schemeSpecificPart
+                    .substringBefore("!")
```
OptimumCode, line 120: `!` is [an allowed character](https://learn.microsoft.com/en-us/windows/win32/fileio/naming-a-file) for file name in Windows. Because of that it looks like this code might not work correctly if there is a `!` in the file path

Nikita-Smirnov-Exactpro:
I have refactored URL.toPath() method
@molotgor could you please update common version to the last special build and try to execute gradlew.bat clean build test on windows

src/main/kotlin/com/exactpro/th2/common/schema/factory/ExactproMetaInf.kt
```diff
@@ -112,15 +112,10 @@ internal class ExactproMetaInf(
...
-            // the code below are added to provide windows compatibility
-            when (scheme) {
-                "file" -> Path.of(this)
-                "jar" -> Path.of(schemeSpecificPart
-                    .substringBefore("!")
-                    .removePrefix("file:"))
-                else -> error("The '$scheme' schema of '$this' URI can't be handled")
-            }
+        internal fun URL.toPath(): Path = when (protocol) {
+            "jar" -> URL(file).toPath() // the code below are added to provide windows compatibility
```
OptimumCode, line 116: note: the code is not below...

OptimumCode (approved):

.github/workflows/build-sanpshot.yml
```diff
@@ -10,6 +10,25 @@ on:
...
+    runs-on: windows-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+      - name: Set up JDK 11 ('zulu')
+        uses: actions/setup-java@v4
+        with:
+          distribution: 'zulu'
+          java-version: '11'
+      - name: Build and test with Gradle
```
OptimumCode, line 23: suggestion: We should add a gradle setup action before we build. In this case, we won't need to download the wrapper every time

```yaml
    - name: Setup Gradle
      uses: gradle/actions/setup-gradle@v4
      with:
        gradle-version: wrapper
```

